PR #15, opened 2024-12-16
14 user friendly installation

krystophny:


krystophny:
@GeorgGrassler this is now completed to merge. It is a more radical change that unifies the build of `QL` and `PAR` versions, dependencies, tools and docs with a single `make` command. I saw no other way. Please try with Tabea on her system. Then continue with #16.

PR #20, opened 2025-01-25
Static linking and dependency management for libneo

krystophny:


GeorgGrassler:
Benchmark 2016 NTV Controlled Fusion could be reproduced, and builds successfully on ITP machines (CODE and no CODE). Awaiting feedback from `MEPHIT` side, then ready to merge with `libneo` branch

PR #22, opened 2025-03-15
Profile

krystophny:
Profiling and fixing performance bottlenecks

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,75 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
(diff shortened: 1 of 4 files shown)


GeorgGrassler:
@krystophny Just so that you aware, there are plans to scratch the splining over radius and instead go with a local lagrange interpolation. The way `neo_magfie.f90` is currently used in `NEO-2` is highly inefficient.

krystophny:
@GeorgGrassler fine. Don't forget the issues with oscillations so choose lowest order (linear) if you don't need more. otherwise, the `spline` library in itpplasma gives an example how to do 1D splines extremely fast and I added the method here for testing it. LAPACK/BLAS tridiaginal solver essentially. Also, `libneo interpolate` and jacobwilliams `bspline` are fast.

PR #28, opened 2025-04-17
Formatting in preparation of refactoring ripple solver

krystophny:


GeorgGrassler (approved):
run `es_0p50824` AUG golden record and reproduced results

**vim -d on the outputs of comparison runs: no difference**
image

PR #56, opened 2025-08-19
Update Readme.md

GeorgGrassler:


GeorgGrassler:
@krystophny I fixed the build issue, it really was just an unnecessary import of sth now not any longer available in the odeint modul. Here now observations about the golden record test:

Stellarator version (full collision operator) comparison against results produced locally on one of our machines (legacy check):
image
It fails with >10^-8 difference in the k_coef, but not that the threshold for the legacy check is about 10^-8. I.e. as we are comparing two runs on two different machines (github vs. or local machines), the expected match is about this regime.

Stellarator version (Lorentz model) comparison against latest stable release (run on same machine)
image
Difference is noticable but stays below 1e-12 relative. Biggest differences (1e-13) belong to attributes of internal magnetic quantities (propagators, ripples, periods, ...), and not usual output results like transport coefficients (about 1e-14/15 difference).

I adapt the test, to check internal magnetics less strictly i.e. with tolerance 1e-12.

krystophny:
PS: Also the D11, D12 seem to be off up to 1e-11 level. I change the tolerance there because those are late outputs with lots of floating point operations in-between (think N*eps with N = 100.000 and original eps=1e-16)

GeorgGrassler:
> PS: Also the D11, D12 seem to be off up to 1e-11 level. I change the tolerance there because those are late outputs with lots of floating point operations in-between (think N*eps with N = 100.000 and original eps=1e-16)

Seems to be more like 1e-10 error, have increased further. These tests should now pass. To counter-act this notion of "laxing" the testing conditions I have added also a value comparison in the performance intensive test of similar tolerance 1e-11/1e-10. Lets see if this also stays inside these bigger bounds.

PR #66, opened 2026-03-27
Move MyMPILib into NEO-2

krystophny:
## Summary

Move `MyMPILib` into `NEO-2` and keep the PR scoped to that change.

## What Changed

- add `COMMON/MyMPILib` as the in-repo home of the library
- move the `MyMPILib` user doc into `DOC/UserDoc/MyMPILib.tex`
- move the dedicated regression test into `TEST/mympilib/`
- wire `COMMON` against the local `NEO2MyMPILib` target instead of `LIBNEO::MyMPILib`
- restore HTTPS-based dependency fetches in `cmake/Util.cmake` and `TEST/cmake/test_libneo_tag.sh`
- remove the extra `LIBNEO_SOURCE_DIR` and `FGSL_PATH` wrapper/config plumbing from this PR
- add `make test` to the active PR workflow so the ported `TEST/mympilib` path runs in CI
- keep a narrow top-level Python interpreter hint so fetched `libneo` configures reliably with the system Python it still requires today

## Why

`MyMPILib` is `NEO-2`-specific code. Keeping it in `libneo` made `libneo` carry an unnecessary MPI build surface and made ownership unclear.

## Impact

- `NEO-2` now owns its MPI abstraction directly
- the CI path keeps using HTTPS for dependency fetches
- the ported `TEST/mympilib` test is now exercised by the active PR workflow

## Verification

```bash
$ make clean
rm -rf build

$ make
...
[438/438] Linking Fortran executable NEO-2-QL/neo_2_ql.x

$ make test
cmake --build --preset default
[1/2] cd /home/ert/code/NEO-2/COMMON/MyMPILib && /home/ert/code/NEO-2/COMMON/MyMPILib/Scripts/do_versioning.sh
Versioning MyMPILib...
[2/2] cd /home/ert/code/NEO-2/build/COMMON && /home/ert/code/NEO-2/COMMON/../ShellScripts/tag_version.sh
...
Test project /home/ert/code/NEO-2/build/TEST
    Start 1: nrutil_test
1/2 Test #1: nrutil_test ......................   Passed    0.07 sec
    Start 2: test_mympilib
2/2 Test #2: test_mympilib ....................   Passed    0.05 sec

100% tests passed, 0 tests failed out of 2
```

## Related

- Companion libneo PR removes `MyMPILib` from `libneo` entirely.
- Related issue: #258


DOC/UserDoc/MyMPILib.tex
GeorgGrassler, line 1: If there is no existing user documentation, and the information is not confirmed, see e.g.

```
\begin{itemize}
  \item Add checkpointing? Could be based on unit of work, maybe needs
  to be implemented in specific code instead.
\end{itemize}
```

adding documentation should be avoided. No information is prefered to wrong information that could lead to wrong assumptions.

cmake/Util.cmake
GeorgGrassler, line 1: why was this changed here? NEO-2 still depends on libneo as before, only MPILib was relocated. Lets fix one thing at the time. And only focus on relocating the MPI Lib as a part of NEO-2 common

GeorgGrassler:
@krystophny I took a view at it: The breaking of the golden record run comes most likely from the changed `find_or_fetch` function workflow. Now it requires ssh authentification in the ci workflow, whereas previously it was find with https. In general, I recommend not fixing everything in one single pr now and just focus on the main objective: moving MPILib into `NEO-2`. So my review:

- avoid drive-by fixes for now if not needed to do the pr
- only move MPILib to NEO-2 and fix the local build process to as was already done in the pr with `add_subdirectory` etc.
- add a test: the golden record test only runs in serial mode. It offers therefore no protection of the correct working of MPILib. 
- add the ported over tests in `TEST` to the CI pipeline

I looked at `MPILib` in `libneo` myself again and it is already quite self contained. There should be no changes necessary at all. Ran a simple recursive diff and got the following back

[differences.txt](https://github.com/user-attachments/files/26354174/differences.txt)

krystophny:
Addressed the review points in b53dc8c.

- Restored HTTPS-based fetch/clone paths in `cmake/Util.cmake` and `TEST/cmake/test_libneo_tag.sh`, so this PR no longer switches CI over to SSH auth.
- Removed the extra `LIBNEO_SOURCE_DIR` and `FGSL_PATH` wrapper/config plumbing again to keep the PR focused on moving `MyMPILib` into `NEO-2`.
- Kept only one narrow build fix: a top-level Python interpreter hint in `CMakeLists.txt`, because fetched `libneo` still configures Python today and CMake was otherwise selecting a broken user-local interpreter here.
- The ported `TEST/mympilib` test stays under `TEST/`, and I added `make test` to the active PR workflow so it is exercised in CI instead of only existing on disk.

Local verification:

```bash
make clean
make
make test
```

Result:

```text
100% tests passed, 0 tests failed out of 2
```

GeorgGrassler:
>     * Kept only one narrow build fix: a top-level Python interpreter hint in `CMakeLists.txt`, because fetched `libneo` still configures Python today and CMake was otherwise selecting a broken user-local interpreter here.

Where was it selecting a broken user-local interpreter? I could not find anything about that in the CMakeConfigs? Or was it doing that in `libneo` configure/build instructions? Cause then this should belong to the libneo fix and not on the consumer side.

@krystophny added also some more comments/questions directly add the changed code lines. That would need clarification first, before approval.

COMMON/CMakeLists.txt
```diff
@@ -95,6 +95,10 @@ add_library(common STATIC
   ${FGSL_USERINTERFACE_SOURCE_FILES}
   )
 add_dependencies(common FGSL)
+target_include_directories(common PUBLIC
```
GeorgGrassler: reason for this target_include_directories change?

CMakeLists.txt
```diff
@@ -6,6 +6,13 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake)
 enable_language(C Fortran)
 enable_testing()
 
+if(NOT DEFINED Python_EXECUTABLE AND EXISTS "/usr/bin/python3")
```
GeorgGrassler: why is this explicit interpreter hint needed now? The import of MyMPIib should not have added any python dependancies.

krystophny:
Removed the hardcoded Python interpreter path in 10d7499. The root cause was a stale uv-managed Python 3.12 symlink in `~/.local/bin/` that lacked dev headers and NumPy — CMake picked it up before the system Python 3.14. Fixed on the local system by removing the broken symlink; the `find_package(Python ...)` call finds the correct interpreter on its own now.

Regarding the inline comments: I can see the reference to source-level comments but the GitHub API returns none — they may still be in pending/draft state. Could you submit the review so I can see and address them?

GeorgGrassler:
> Regarding the inline comments: I can see the reference to source-level comments but the GitHub API returns none — they may still be in pending/draft state. Could you submit the review so I can see and address them?

Jep, sorry about that. Some of it is already resolved now I think, take a look.

krystophny:
Addressed the remaining scope concerns in `824ea2b`.

Changes removed again:
- top-level `find_package(Python ...)` from `NEO-2/CMakeLists.txt`
- the extra `target_include_directories(common ...)` change in `COMMON/CMakeLists.txt`
- speculative `DOC/UserDoc/MyMPILib.tex` and its inclusion from the user manual

That keeps this PR focused on relocating `MyMPILib` into `NEO-2` and adding the ported test coverage, without introducing a new consumer-side Python dependency or unrelated documentation/config changes.

Local verification on a fresh build dir:

```bash
make clean
make test
```

Result:

```text
100% tests passed, 0 tests failed out of 2
```

krystophny:
@GeorgGrassler latest cleanup is now on the branch in `824ea2b`.

Latest changes:
- removed the top-level Python requirement from `NEO-2/CMakeLists.txt`
- removed the extra `target_include_directories(common ...)` change
- removed the speculative `MyMPILib` user-doc file and its include
- re-ran `make clean && make test` locally: `100% tests passed, 0 tests failed out of 2`

Does this address the remaining scope concerns on the NEO-2 side?

krystophny:
Status (2026-05-18): still in progress; lands paired with itpplasma/libneo#259, targeting this summer.

Issue #69, opened 2026-04-13
Support Om_tE as independent profile input for NEO-2-QL


krystophny:
## Goal

Allow the ExB rotation frequency Om_tE (rad/s) to be specified as an independent radial profile input for the non-axisymmetric stage of NEO-2-QL, instead of being forced to use the neoclassically computed value from the axisymmetric run.

## Background

Currently, the radial electric field enters NEO-2-QL through `MtOvR` (toroidal Mach number / major radius), which is species-dependent. The code either:
- `isw_calc_Er=1`: computes Er neoclassically from axisymmetric transport coefficients, derives per-species `MtOvR_spec`
- `isw_calc_Er=0`: uses a single scalar `MtOvR` from the namelist -- physically wrong for multi-species (same MtOvR for all species)

`Om_tE` (the ExB drift frequency) is species-independent and is the natural physical input quantity. The conversion is:
```
MtOvR_spec(a) = Om_tE / sqrt(2 * T_a / m_a)
```

## Checklist

- [x] Extract pure conversion functions `Om_tE <-> MtOvR_spec` into `NEO-2-QL/er_rotation_mod.f90` with unit tests
- [x] Wire conversion into `compute_Er`: replace inline formula in `ntv_mod.f90` with two-step via `er_rotation_mod`. Compute `Om_tE` as intermediate.
- [x] Add `Om_tE` as module variable in `ntv_mod.f90` alongside `MtOvR`. Store when `compute_Er` runs. Output to HDF5.
- [x] Add `Om_tE_prof` to HDF5 multispec input: read `Om_tE_prof(num_radial_pts)` in `neo2.f90` (optional field, backward compatible). In `prepare_mulitspecies_scan`, set per-radial-point `Om_tE`.
- [x] Add `isw_calc_Er=2` mode: new branch in ripple solver -- use externally specified Om_tE, convert to species-specific `MtOvR_spec` via `Om_tE_to_MtOvR_spec()`.
- [x] Update Python tooling: add Om_tE profile to `generate_multispec_input.py` HDF5 output. Add tests.
- [x] Add Om_tE to HDF5 output in `write_multispec_output`.
- [ ] Consider deprecating global `MtOvR` scalar: replace with per-species array + `Om_tE`. Tracking issue.

## isw_calc_Er values

| Value | Behavior |
|-------|----------|
| 0 | Use scalar MtOvR from namelist (legacy, single-species only) |
[...]

PR #71, opened 2026-04-13
Add er_rotation_mod with Om_tE to MtOvR conversion functions

krystophny:
## Summary

- Add pure conversion functions between species-independent Om_tE (rad/s) and species-dependent MtOvR_spec in `COMMON/er_rotation_mod.f90`
- Add unit tests for round-trip, known values, multi-species consistency, and consistency checking
- First step toward #69 (Om_tE as independent profile input)

## Test plan

- [x] `make` builds successfully
- [x] `make test` passes 100% (2/2 unit tests)
- [ ] CI golden record tests pass (no production code changed, purely additive)


krystophny:
Added minimal guards in ba136d8: early return for empty arrays in `MtOvR_spec_to_Om_tE` and size-mismatch check in `check_Om_tE_consistency`.

The remaining suggestions (error codes, removing `pure`, `assert_eqn`, NaN returns) are overly defensive for internal functions -- `T_spec`, `m_spec`, and `MtOvR_spec` are always allocated with the same size from `collisionality_mod`, and `num_spec >= 1` is guaranteed by the input validation in `neo2.f90`.

PR #79, opened 2026-04-14
Add transport k golden record test to CI

### LINKED ISSUE #77: Add repo-local regression for transport k vs Sauter reference
## Summary

Add a low-churn regression that validates the NEO-2 transport coefficient ratio

```text
k_transport = 5/2 - D32_ii / D31_ii
```

against a committed Sauter reference on a small axisymmetric / AUG-derived fixture, without requiring external GACODE/NEO binaries at test time.

## Motivation

In the recent AUG 30835 audit, a previous comparison mistake treated `Vtht_prof_spec` as a direct `k dT/dr` object. After tracing `NEO-2-QL/ntv_mod.f90`, that interpretation was corrected: `Vtht_prof_spec` is a derived Boozer contravariant profile built from the full force-balance / transport machinery.

Using the transport definition instead,

```text
k = 5/2 - D32_ii / D31_ii
```

we found that NEO-2 `k` is very close to the Sauter `k` curve on the audited AUG case:

- median `|Sauter - NEO-2|` ~= `3.1e-3`
- max `|Sauter - NEO-2|` ~= `6.9e-3`

That makes `k_transport` a good candidate for a stable regression target.

## Goal

Create the smallest possible test/fixture addition that protects this result without pulling external-code runtime dependencies into CI.

## Preferred strategy (minimum churn)

1. **Do not add a new Fortran-side test harness.**
2. Reuse the existing Python regression style in `python/test/test_compute_omte.py`.
3. Add a tiny helper that extracts the ion-ion block and computes
   - `k_transport = 2.5 - D32_ii / D31_ii`
4. Compare that curve against a **committed reference fixture**:
   - either extend an existing repo-local fixture with a stored Sauter reference curve
   - or add a small new fixture (`.npz` / `.h5`) containing only the needed radial grid and Sauter `k`
5. Keep fixture regeneration as an **offline developer script**, not part of CI.

## Why this is preferable

- no runtime dependency on external GACODE/NEO binaries in CI
- no new coupling into the main Fortran executable
- matches the existing repo pattern: committed HDF5/NPZ fixtures + Python regression checks
- isolates the tested semantic claim to the actual transport coefficients that define NEO-2 `k`

### Option A (preferred)

Add one new Python regression test using an existing / lightly extended fixture:

- input: repo-local `neo2_multispecies_out.h5` or AUG `.npz` fixture
- compute `k_transport` from `D31_AX`/`D32_AX`
- compare to stored Sauter reference values with a tolerance based on the current audit

This is the smallest CI-facing change.

### Option B

Extend the golden-record postprocessing to archive `D31_AX`, `D32_AX`, and derived `k_transport`, then compare those against committed expected outputs.

This is still reasonable, but larger in scope than Option A and mixes the semantic validation into the heavier golden-record workflow.

### Option C (not preferred initially)

Add new detailed output directly from the Fortran main solely to support `k` extraction.

Probably unnecessary, since the needed transport coefficients are already in the output.

## Open design questions

- Which fixture should be the canonical source for this validation?
  - existing axisymmetric HDF5 fixture
  - existing AUG reference NPZ fixture
  - a new compact fixture carrying only `rho`, `D31`, `D32`, `k_sauter_ref`
- Should the reference be a full radial profile or a few representative surfaces?
- What tolerance is appropriate to remain stable but still meaningful?

## Acceptance criteria

- CI runs a repo-local test with no external-code runtime dependency
- the test derives `k_transport` from stored NEO-2 coefficients
- the test compares against a committed Sauter reference curve / points
- the test documentation makes explicit that `Vtht_prof_spec` is **not** the object from which `k` should be inferred

krystophny:
## Summary

- Add `make transport_k` to both CI golden record test steps (stable and main comparison)
- The test case lives in the data repo: `TESTS/NEO-2/golden_record/transport_k/`
- Single AUG flux surface (s=0.32151), reduced .bc file (5 surfaces), LAG=3 for speed (~3 min with 2 MPI ranks)
- Full Sauter comparison and gacode NEO provenance documented in data repo

## What lives where

**NEO-2 repo (this PR)**: 6 lines in `.github/workflows/test-on-pr.yml`

**Data repo (already pushed to main)**:
- `TESTS/NEO-2/golden_record/transport_k/` - golden record test case (Makefile, reduced .bc, neo2.in, reference summary.h5)
- `AUG/NEO-2/30835/.../gacode_neo_comparison/` - full gacode NEO comparison, Sauter CSV, traceability stages 0-6

Closes #77

## Test plan
- [ ] CI runs `make transport_k` successfully
- [ ] Reference and current summary.h5 match within tolerance


krystophny:
Merged main. Root cause of the run-golden-record failure was on the data side: the transport_k test case had been removed from the data repo main during the AUG 30835 data move, while the golden_record Makefile still referenced it. Restored it there (cfdacdd2) with two harness fixes: executable bit on reference/run_neo2.sh (the April failure) and run_neo2.sh added to the .run prerequisites in current/Makefile. Verified locally: reference and current runs complete and compare_h5.py passes at 1e-15.

PR #86, opened 2026-05-28
Resolve libneo via LIBNEO_REF override, default to release branch

krystophny:
Resolves the libneo dependency ref with this precedence:

1. `LIBNEO_REF` (cache or env) - a branch, tag or commit, validated against the remote; ignored if absent instead of failing the fetch. An upstream libneo release sets this to build NEO-2 against a candidate.
2. the committed `LIBNEO_RELEASE` (`release/26.2`) - the libneo release branch this code tracks. Never libneo main.

Renames the override from `LIBNEO_BRANCH` to `LIBNEO_REF` since it accepts any git ref. Companion CI change adds the matching `workflow_dispatch` trigger.

### DIFF
--- a/COMMON/ProjectConfig.cmake.in
+++ b/COMMON/ProjectConfig.cmake.in
@@ -2,6 +2,7 @@
 set (CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR})
 
 include(Util)
+set(LIBNEO_RELEASE release/26.2 CACHE STRING "libneo release branch to track")
 find_or_fetch(libneo)
 
 ### Path to the external libraries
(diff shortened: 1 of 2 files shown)


krystophny:
@GeorgGrassler heads-up: this PR is part of a cross-repo change unifying how the codes resolve libneo (a validated `LIBNEO_REF` override and tracking the `release/26.2` branch by default), plus an external CI trigger so the libneo release gate can build this code against a candidate ref. It needed a few more changes than expected, so your feedback would be valuable before it lands. We plan to merge the whole set on Monday, 2026-06-15. Companion CI PR: #87.

krystophny:
Not needed: NEO-2's resolver already honors LIBNEO_GIT_TAG (forwarded by the Makefile), so the release gate drives it via #87 alone. Avoids the resolver rewrite and keeps the $CODE local-path shortcut.

PR #87, opened 2026-06-08
Allow external libneo_ref trigger in CI

krystophny:
Wires NEO-2 into libneo's per-PR reverse-dependency gate (libneo branch feat/release-downstream-gate). NEO-2 is a fast-only downstream in the gate manifest: the gate dispatches this workflow with `-f libneo_ref=<sha>` and never passes `full`.

This PR implements that dispatch contract:

- `unit-tests.yml` (new): the workflow the gate dispatches. It must live on `main` to be dispatchable. `workflow_dispatch` accepts `libneo_ref`. The build step exports it as `LIBNEO_GIT_TAG`; the Makefile forwards `-DLIBNEO_GIT_TAG=` to cmake, and `find_or_fetch(libneo)` in `cmake/Util.cmake` uses it as the FetchContent `GIT_TAG`, so NEO-2 builds against the candidate libneo. On push and PR `libneo_ref` is empty and the pinned libneo resolves as before.
- `make test` runs `ctest --test-dir TEST`, which covers only the fast unit tier (nrutil, mympilib, er_rotation; 30 s timeouts). The golden-record and PAR/QL reconstruction suites live in `NEO-2-PAR`/`NEO-2-QL` and stay in `test-on-pr.yml` on push/PR only, so a gate dispatch runs the sub-minute set.

Normal push/PR runs are unchanged.

## Hermetic dependency fetching

Dependency resolution no longer reads `$ENV{CODE}`, so the build is hermetic regardless of the ambient environment:

- `cmake/Util.cmake`: `find_or_fetch` dropped the `$ENV{CODE}/<dep>` bypass. libneo (and any other dep) is now always fetched via FetchContent by default. A local checkout is selected only through the explicit `-D<DEP>_SOURCE_DIR=<path>` cache option, or its alias `-D<DEP>_PATH=<path>` (e.g. `-DLIBNEO_PATH=`). The `-DLIBNEO_REF=` pin path is unchanged.
- `COMMON/ProjectConfig.cmake.in`: FGSL dropped the `$ENV{CODE}/external/fgsl-1.6.0/` prebuilt-path branch. FGSL builds via the bundled ExternalProject by default; `-DFGSL_PATH=<dir>` selects a local prebuilt FGSL. FGSL remains a live dependency on this branch; its removal is owned by the fortnum/drop-gsl migration, not this PR.

Verification: `rg -i '\$ENV\{CODE\}'` over the cmake sources returns nothing. A fresh `cmake -S . -B <dir> -G Ninja` with `CODE` exported and a decoy `$CODE/libneo` present logs `Using libneo branch main from https://github.com/itpplasma/libneo.git` and populates `_deps/libneo-src` from the remote, confirming `$CODE` is ignored. `-DLIBNEO_PATH=<path>` switches to the local checkout (`Using local libneo in <path>`).

**Merge order: merge this PR first, then itpplasma/libneo#278.** `unit-tests.yml` must exist on NEO-2 `main` before libneo's gate can dispatch it. This PR is otherwise standalone (the workflow runs on its own push/PR), so it is safe to merge first.


GeorgGrassler (changes requested):
@krystophny might I really advice to split the workflows in this case? The whole `test-on-pr.yml` is basically skip for your main use case. Could we not make an extra workflow just for the unit tests, that is always triggered? Would be in general an improvement and maybe easier to trigger from libneo. Then you can maybe skip the slow workflow completely and dont need to make those complicated combinations of triggers for each "long" job.

krystophny:
Done, split as you suggested.

- `unit-tests.yml` (new): the always-on fast tier. Builds and runs `make test` on push, PR, and `workflow_dispatch`, accepting `libneo_ref`. This is what the libneo gate triggers.
- `test-on-pr.yml`: back to its prior form. Golden-record, performance, and PAR run on push/PR only. The `full` input and the per-step dispatch conditionals are gone.

libneo side (itpplasma/libneo#278): the gate now dispatches `unit-tests.yml`, and `ci/downstreams` got a fourth `full` column. NEO-2 is `full=no`, so the gate never sends it `-f full=true`, and its golden-record/PAR suite stays on NEO-2 PRs instead of the gate.

krystophny:
Merge order: this PR first, then itpplasma/libneo#278. `unit-tests.yml` must be on NEO-2 `main` before libneo's gate can dispatch it. This PR is standalone otherwise, so it is safe to merge first.

GeorgGrassler:
@krystophny is the `make test` still part of the big CI? Should it not be removed as it is redundant with this new workflow?

krystophny:
Yes, the redundant `make test` in `test-on-pr.yml` (the Run unit tests step at lines 104-106) was removed in this PR. The fast tier now lives exclusively in `unit-tests.yml`; `test-on-pr.yml` runs only the golden-record and PAR/QL suite.

GeorgGrassler:
@krystophny what is the indent behind the last three commits? I though this was all done?

cmake/Util.cmake
```diff
@@ -1,17 +1,25 @@
 include(FetchContent)
 
 function(find_or_fetch DEPENDENCY)
-    if(DEFINED ENV{CODE} AND EXISTS $ENV{CODE}/${DEPENDENCY})
-        set(${DEPENDENCY}_SOURCE_DIR $ENV{CODE}/${DEPENDENCY})
-        message(STATUS "Using ${DEPENDENCY} in $ENV{CODE}/${DEPENDENCY}")
+    # Explicit local override only. Pass -D<DEP>_SOURCE_DIR=<path> (or its
+    # alias -D<DEP>_PATH=<path>) on the cmake command line to build against a
```
GeorgGrassler: why is there an alias even necessary?

krystophny:
Right, the split was done at `56cfa1c`. The three later commits just make the gate reproducible and fix one break:

- `9225abc`: rename `LIBNEO_GIT_TAG` -> `LIBNEO_REF` and `unexport` it in the Makefile so ambient shell env can't leak into cmake. Also removes the redundant `make test` you flagged.
- `1a877a4`: drop the `$ENV{CODE}` bypass so a local `$CODE/libneo` can't shadow the ref the gate hands us. Local checkout now needs explicit `-DLIBNEO_PATH=`.
- `8a81cad`: the reference build runs the last NEO-2 release, whose old Makefile only knows `LIBNEO_GIT_TAG`; passing only `LIBNEO_REF` left libneo on `main` (where `LIBNEO::MyMPILib` is gone) and broke it. Pass both names so the pin holds.

Done now.

cmake/Util.cmake
```diff
@@ -1,17 +1,25 @@
...
+    # local checkout. With neither set, the dependency is always fetched, so
+    # the build is hermetic regardless of any ambient environment.
+    string(TOUPPER ${DEPENDENCY} DEPENDENCY_UPPER)
```
GeorgGrassler, line 5: What is the point of DEPENDENCY_UPPER vs DEPENDENCY?

Makefile
```diff
@@ -2,11 +2,15 @@ BUILD_DIR := build
 BUILD_NINJA := $(BUILD_DIR)/build.ninja
 CONFIG ?= Release
 
+# Prevent ambient shell values from silently changing the libneo fetch.
+# Pass LIBNEO_REF=<ref> on the make command line to override explicitly.
+unexport LIBNEO_REF LIBNEO_PATH LIBNEO_GIT_TAG LIBNEO_BRANCH
```
GeorgGrassler: are all those envs even supported in the current version? Like the LIBNEO_GIT_TAG is not present in the code base anymore.

.github/workflows/test-on-pr.yml
```diff
@@ -123,9 +119,11 @@ jobs:
           echo "Building reference version from latest stable release: $LATEST_RELEASE"
           git checkout $LATEST_RELEASE
 
-          # Use same tag for libneo (synchronized versioning)
+          # Use same tag for libneo (synchronized versioning).
+          # The released Makefile honors LIBNEO_GIT_TAG; current code honors
+          # LIBNEO_REF. Pass both so the pin works regardless of release age.
           echo "Using synchronized libneo tag: $LATEST_RELEASE"
-          make LIBNEO_GIT_TAG=$LATEST_RELEASE
+          make LIBNEO_GIT_TAG=$LATEST_RELEASE LIBNEO_REF=$LATEST_RELEASE
```
GeorgGrassler: current version only honors LIBNEO_REF, so why not have the workflow (Makefile, CI, ...) honor **only** this as well? For the reference build (stable release) you just pass LIBNEO_GIT_TAG, specifically because you honor its legacy behavior/Makefile. Their workflow only knows LIBNEO_GIT_TAG. If you want to avoid silently failure once we update the stable release past this current merge i.e. after we changed the name, you should just place a loud guardrail in the Makefile that rejects any passes of LIBNEO_GIT_TAG i.e. "LIBNEO_GIT_TAG is deprecated, use LIBNEO_REF instead!" This makes it that before name change only sees the old name (we do not temper with that old Makeprocess) and the current one only accepts the new version. Here we only want to test against the **latest** stable release, so I see no reason why we would keep the pinning longer viable until the next stable release anyway, and this ensures that we clean up after ourselves.

GeorgGrassler (changes requested):

cmake/Util.cmake
krystophny: It is not. Dropped the alias in 2f62a3c; a local checkout is now selected only via -D<dep>_SOURCE_DIR (the variable FetchContent itself populates).

cmake/Util.cmake
krystophny: The cache vars are upper-case by convention (LIBNEO_REF), but the dependency name is lower-case because it is also the repo path (itpplasma/libneo.git) and the FetchContent target. TOUPPER bridges the two so one find_or_fetch call serves any dep.

Makefile
krystophny: No. LIBNEO_GIT_TAG/LIBNEO_BRANCH are gone from the current tree. In 2f62a3c the Makefile honors only LIBNEO_REF (and ignores an ambient one).

.github/workflows/test-on-pr.yml
krystophny: Done in 2f62a3c. Current code honors only LIBNEO_REF; the reference build passes only LIBNEO_GIT_TAG, which the old released Makefile needs. We just dropped the old name from current code rather than adding a deprecation $(error), so there is nothing to guard.

GeorgGrassler: My point was that you should keep an error gate, so that when we update the stable release that is currently used as reference, we hear it fail loudly so that we update the CI then.

GeorgGrassler (changes requested):

.github/workflows/test-on-pr.yml
krystophny: Done. The current Makefile now errors on `LIBNEO_GIT_TAG` and `LIBNEO_BRANCH` and points to `LIBNEO_REF`. The reference build still uses the old released Makefile after checkout.

GeorgGrassler (approved):

PR #92, opened 2026-06-14
quadpack: back gsl_integration_routines_mod by fortnum, drop unused gsl_roots

krystophny:
## Scope

Re-express the QUADPACK-pattern 1D integrators in `gsl_integration_routines_mod` on fortnum, and remove the unused `gsl_roots_routines_mod`. `fint1d_qag/qags/qagp/qagiu` map to `integrate_qag/qags/qagp/qagiu`; `fint1d_cquad` runs on the adaptive QAGS path (bisection plus Wynn-epsilon extrapolation, which covers the endpoint singularities CQUAD targets). The user integrand reaches fortnum through a host-associated `panel(x, ctx)` wrapper, and `fortnum_status_t` feeds the existing `disp_gsl_integration_error` accumulator. The public interface (module name, generic procedure names, `result(2)` value/abserr return) is preserved so `COMMON/collop_compute.f90` and both `collop_definitions.f90` compile unchanged. The internal kind alias is renamed `dp` -> `wp` so the module no longer leaks a `dp` that clashes with the callers' `nrtype` `dp` on use-association. `gsl_roots_routines_mod` has no call sites in the tree and is deleted together with its build-list entry.

## Dependency removed

GSL/FGSL QUADPACK integration (`gsl_integration_*`, `fgsl_function`, the GSL workspace handles) and the GSL one-dimensional root solver (`gsl_root_*`).

## Verification

fortnum built at `a7faa3c` via `fo build` (exit 0). The migrated integration module compiled standalone against the fortnum module path:

```
$ gfortran -c -ffree-line-length-none -I$CODE/fortnum/build/include \
    COMMON/gsl_integration_routines_mod.f90
exit: 0
```

The object references only fortnum, no GSL/FGSL:

```
$ nm gsl_integration_routines_mod.o | grep -iE " U " | grep -iE "gsl|fgsl|amos|slatec|fftw"
(no output)
$ nm gsl_integration_routines_mod.o | grep -ic fortnum
4
```

No remaining caller of the deleted root solver:

```
$ rg -l gsl_roots_routines_mod --glob '!COMMON/gsl_roots_routines_mod.f90'
(no output)
```

## Unverified

Full NEO-2 executable link is not exercised in this PR (it lands with the CMake wiring at the tip of the stack). The local `libneo` checkout cannot configure NEO-2 here because `libneo`'s `src/magfie/CMakeLists.txt` links the `FFTW::DoubleThreads` target, which is not created when `libneo` is consumed via `add_subdirectory` in this environment (present on `libneo` main, independent of this change). Golden-record physics tests run in the project's GitHub Actions, not locally.


krystophny:
Replaced by the reordered fortnum migration stack: P1 #95 (add fortnum, base main), P2 #96 (besselk), P3 #97 (quadpack, this PR's content), P4 #98 (bspline), P5 #99 (drop GSL/FGSL). quadpack now lives in #97.

PR #97, opened 2026-06-14
P3 quadpack: gsl_integration_routines_mod -> fortnum, drop unused gsl_roots

krystophny:
## Merge order

Third PR of the reordered fortnum migration stack. Base: P2 (#96).

Re-expresses the QUADPACK-pattern 1D integrators on fortnum: fint1d_qag/qags/qagp/qagiu map to integrate_qag/qags/qagp/qagiu, and fint1d_cquad runs on the adaptive QAGS path. The user integrand reaches fortnum through a host-associated panel(x, ctx) wrapper, and fortnum_status_t feeds the existing disp_gsl_integration_error accumulator. The public interface (module, generic names, result(2) value/abserr) is preserved so collop_compute.f90 and both collop_definitions.f90 compile unchanged; the internal kind alias is renamed dp -> wp to avoid clashing with the callers' nrtype dp. gsl_roots_routines_mod has no call sites and is removed with its build-list entry.

```
```


krystophny:
Merged main; declared fortnum before libneo so this repo's pin wins, and pinned lazy-fortran/fortnum#57. Root cause of the 6 h golden timeout: fortnum was fetched at libneo's pin and compiled under libneo's -ffast-math scope, which removed its NaN handling; collision-operator integrands hitting 0*inf at isolated nodes filled the matrices with NaN and zhseqr failed on every Arnoldi iteration. The pinned fortnum adds a coefficient-difference cquad error estimate, zeroes non-finite samples, and forces -fno-fast-math on its target; the QL performance case now completes locally with zero NaN integrals and healthy convergence (4 bad modes, 17 stabilized iterations). Unit tests pass locally (5/5). The lorentz exact-vs-main compare still reports the expected backend-swap residual (denmm 4.9e-10, ailmm 2.5e-10), which is the GSL reference's own stopping error at epsabs=1e-11.

krystophny:
Status of the golden-record investigation:

- Bisect across the stack: #95 (wiring) and #96 (besselk) pass the golden record; failures start here (#97). The #98 and #99 runs show bit-identical differences to #97, so the 1D-integrator swap is the sole source, bspline adds nothing.
- Only two integrator entry points are live in collop_compute: fint1d_cquad (finite segments between basis knots) and fint1d_qagiu (Maxwellian tails). Everything else has no call sites.
- The ql exact-check failure is reproduced locally bit-for-bit (ailmm_aa 3.567174273426909e-08, identical to CI), so the discrepancy is deterministic.
- Now tracing each fint1d call in the ql run with the GSL result as oracle (same call, both backends, per-call CSV) to pin down which integrals and which backend path diverge, before deciding on the fortnum-side fix.

No tolerance changes; the golden record stays as is.

krystophny:
Root cause identified, per-call and per-evaluation tracing complete.

Instrumented the fortnum-backed shim to run every live integrator call through both backends (fortnum and FGSL/GSL as oracle) in the same process during the ql golden-record case: 3.07M calls traced.

- fint1d_qagiu and the 2.76M inner-kernel fint1d_cquad calls agree to <= 1.2e-15 relative. fortnum accuracy is not the problem.
- All divergent calls (1756 of 3.07M) sit on collision-operator segments whose endpoints are B-spline knots. With collop_bspline_order=2 the matrix-element integrands jump exactly at the knots (piecewise-constant basis derivatives), which is why the integrate() wrapper splits at t_vec in the first place.
- Per-evaluation dumps show the actual mechanism: the Clenshaw-Curtis node arithmetic (mid + half*cos) of BOTH backends can land one ulp outside the requested segment, e.g. x = 2.66666666666666607 for a segment starting at a = 8/3 = 2.66666666666666652. That sample sits on the far side of the knot where the integrand is 5.4e5 instead of 0. GSL's CQUAD then subdivides against the phantom kink and keeps a spurious ~2e-10 contribution (its own abserr ~7e-9 admits it); fortnum's subdivision happens to end on nodes that round back onto the knot and returns the exact 0. Every ailmm/denmm difference in the golden record traces back to these one-ulp boundary slivers, amplified downstream into the 1e-6-level D-coefficient diffs.

So the diff is a boundary-sampling artifact of adaptive quadrature on knot-discontinuous integrands, not a backend accuracy regression, and it cannot be closed by making fortnum more accurate (fortnum already returns the exact integral on the offending segments).

Fix under verification: clamp the integrand abscissa to the open segment (one ulp inside each end) in the fint1d_cquad wrappers, so every segment integrand is continuous on its closed interval and both backends resolve it to machine precision. This must land on main as well (the exact check builds its reference from main), which also removes the sliver artifact from the GSL-based results. Local two-sided ql run with the clamp on both builds is in progress; numbers follow.

krystophny:
Clamp fix verified two-sided and pushed; here are the numbers.

The abscissa clamp (evaluate one ulp inside each segment) is now in three places: on main as PR #117 (GSL/FGSL wrappers), and on this stack as 097ee5e (fortnum shim, merged through to the P4 and P5 branches). fortnum pin stays at f849f59 - the tracing shows fortnum needs no accuracy fix: after the clamp, all 3.07M traced integrator calls in the ql case agree with GSL to a few ulp, and fortnum was the side returning the exact value on the previously divergent segments.

Local two-sided golden runs, current (this stack, fortnum) vs reference (clamped main, GSL):

lorentz (compare at 1e-14 / magnetics 1e-12):
- before clamp: ailmm 2.46e-10, denmm 4.86e-10 (and _a/_aa variants)
- after clamp: denmm clean below the gate; only ailmm/ailmm_aa remain, at 8.53e-12

ql (compare at 1e-15):
- before clamp: ailmm_aa 3.57e-08, gamma_out 1.44e-07, D12_AX 2.77e-06
- after clamp: ailmm_aa 7.76e-11, gamma_out 1.47e-09, D12_AX 1.53e-09

What remains is not a quadrature defect on either side: the ailmm assembly sums segment integrals of magnitude ~4e5 that cancel down to elements a few thousand times smaller. Each segment integral agrees across backends to 1-3 ulp (2-7e-16 relative; the irreducible difference between two correct implementations with different node sets), and the cancellation amplifies that to the observed 1e-11-level residual. Closing that would require bit-identical arithmetic with GSL, i.e. porting GSL's CQUAD decision-for-decision, which fortnum's clean-room policy rules out.

Practical consequence: the exact check can only go green for this stack once #117 is on main (so the freshly built reference is also clamped), and even then lorentz/ql will show the 1e-11-class ailmm residual against the 1e-14/1e-15 gates. That residual is the true, fully characterized backend-swap delta - three to four orders smaller than what these PRs showed before.

PR #98, opened 2026-06-14
P4 bspline: gsl_bspline_routines_mod -> fortnum_bspline

krystophny:
## Merge order

Fourth PR of the reordered fortnum migration stack. Base: P3 (#97).

Replaces the FGSL B-spline workspace with the caller-owned fortnum bspline_workspace_t. init_bspline/set_bspline_knots map to bspline_init/bspline_set_knots; basis values come from bspline_eval_basis and the all-orders derivative table from bspline_eval_deriv. The public interface is preserved (init_bspline, set_bspline_knots, bspline_eval, bspline_deriv_eval, the taylorExpansion flag, fgsl_double now aliased to real64) so collop_bspline.f90 compiles unchanged. The dby/dby_end arrays are reindexed 0:order to match fortnum's (0:nderiv, ncoef) layout; the Taylor extrapolation above the last breakpoint stays in NEO-2.

```
```


krystophny:
## ailmm/denmm drift: the fortnum pin predates the QUADPACK quadrature match

The collision-operator integrals reach fortnum through `integrate_a_to_b` (finite QAGS) and `integrate_a_to_infinity` (QAGIU) in `gsl_integration_routines_mod`. This branch pins fortnum to `v0.1.0`, which is older than the fortnum commit that matched the QAGS/QAGP/QAGIU adaptive extrapolation to QUADPACK `dqagse`/`dqagpe`/`dqagie`. On `v0.1.0` the extrapolation is unmatched, which is the most likely origin of the ~8.9e-8 ailmm/denmm drift against the GSL-generated reference (the bspline is bit-identical, so the drift rides the predecessor quadpack pin).

Independent checks against GSL on the relevant integrand classes:
- fortnum `integrate_qagiu(inf=1)` vs `gsl_integration_qagiu` on Chandrasekhar/Maxwell semi-infinite kernels (`x^n exp(-x^2)`): agree to ~1e-14, matching the analytic values.
- the finite QAGS/QAGP path matches QUADPACK `dqagse`/`dqagpe` on the current fortnum main.

### Proposed fix

Re-pin fortnum from `v0.1.0` to the main commit carrying the matched quadrature (branch `reorder/p4-bspline-fortnum-pin`). Local builds that resolve fortnum from a working checkout already use that code; the pin bump aligns CI with it.

Note: the token-gated reconstruction test data was not available in this environment, so the full ailmm/denmm exact-check was not re-run here; the re-pin is the candidate fix backed by the per-integrand quadrature comparisons above. No tolerance was weakened and no reference regenerated.

krystophny:
Merged main; declared fortnum before libneo so this repo's pin wins, and pinned lazy-fortran/fortnum#57 (coefficient-difference cquad error estimate, zeroed non-finite samples, -fno-fast-math defense; see #97 for the diagnosis of the NaN-driven 6 h golden timeout). Build and unit tests pass locally (5/5).

krystophny:
Golden-record root cause and fix are tracked in #97 (see the tracing and post-clamp numbers there). This branch received the same abscissa clamp via merge from the P3 branch; the bspline swap itself contributes zero numerical difference (the P3/P4/P5 golden diffs were bit-identical). Exact-check green additionally requires #117 on main so the reference build is clamped too.

PR #99, opened 2026-06-14
P5 drop-gsl: remove GSL/FGSL now that fortnum backs every module

krystophny:
## Merge order

Final PR of the reordered fortnum migration stack. Base: P4 (#98).

No module uses GSL or FGSL anymore. Removes the FGSL ExternalProject (cmake/fgsl.cmake), find_package(GSL), the FGSL include and link paths, and the add_dependencies(common FGSL) edge. fortnum is the sole numerical backend of the common library.

## Verification

Full reordered stack tip built locally with `$CODE` unset (https FetchContent of fortnum 4c560b0, no GSL): neo_2_par.x and neo_2_ql.x both link; ctest 3/3 pass.

```
```


krystophny:
Merged main; declared fortnum before libneo so this repo's pin wins, and pinned lazy-fortran/fortnum#57 (coefficient-difference cquad error estimate, zeroed non-finite samples, -fno-fast-math defense; see #97 for the diagnosis of the NaN-driven 6 h golden timeout). Build without GSL/FGSL and unit tests pass locally (5/5).

krystophny:
Golden-record root cause and fix are tracked in #97 (see the tracing and post-clamp numbers there). This branch received the same abscissa clamp via merge from the P3 branch; the bspline swap itself contributes zero numerical difference (the P3/P4/P5 golden diffs were bit-identical). Exact-check green additionally requires #117 on main so the reference build is clamped too.

PR #101, opened 2026-06-24
sparse_mod: route solves through fortsparse (out-of-process UMFPACK)

krystophny:
## Summary

`sparse_mod` now routes every sparse solve through a thin wrapper instead of the in-process umf4 C wrappers. The new `COMMON/sparse_solve_fortsparse.f90` exposes two leaf procedures, `fs_solve_real` and `fs_solve_complex`, and the four call-site clusters in `sparse_mod` (real/complex, single- and multi-RHS) call only those. No client code outside `sparse_mod` changes; callers still drive the same factor/solve/free lifecycle through `iopt`.

`neo_2_ql.x` and `neo_2_par.x` no longer link GPL UMFPACK. UMFPACK now runs out-of-process in the fortsparse helper, reached over shared memory. The numerics are unchanged: the same UMFPACK does the factorization and solve, just in a separate process.

The four in-process umf4 C wrappers (`COMMON/umf4_f77*wrapper*.c`) are removed. fortsparse is fetched from the public `lazy-fortran/fortsparse` via `cmake/fortsparse.cmake`. CI gains `libsuperlu-dev`, required by fortsparse.

## Verification

Local build (`make ninja`): green, binaries current.

Unit tests (`ctest --test-dir TEST --output-on-failure`):

```
1/5 Test #1: nrutil_test ........................   Passed    0.07 sec
2/5 Test #2: sparse_solve_fortsparse_test .......   Passed    0.81 sec
3/5 Test #3: sparse_boundary_equivalence_test ...   Passed    0.32 sec
4/5 Test #4: er_rotation_test ...................   Passed    0.06 sec
5/5 Test #5: test_mympilib ......................   Passed    0.06 sec

100% tests passed, 0 tests failed out of 5
```

UMFPACK linkage (`ldd ... | grep -i umfpack`):

```
neo_2_ql.x:  NO umfpack
neo_2_par.x: NO umfpack
fortsparse_umfpack_helper:  libumfpack.so.6 => /usr/lib/libumfpack.so.6
```

The helper links UMFPACK; the NEO-2 executables do not.

## Before marking ready

The existing golden-record and performance CI (`test-on-pr.yml`, TOL 1e-14..1e-15) runs only once this draft is marked ready for review. NEO-2 CI now also needs `libsuperlu-dev` (added in this PR) and fetches the public `lazy-fortran/fortsparse`.


krystophny:
run-golden-record died on the runner mid-QL-run: local PSS profiling shows the IPC backend peaking ~2 GB above the in-process path, dominated by the adapter's csc_t copy per factorization. fortsparse now streams raw CSC arrays straight into the helper's shared mapping (lazy-fortran/fortsparse#1, merged); the adapter here uses that path and drops the copy. Multispecies QL case completes locally (peak 18.5 GB, exit 0); branch also merged with main.

PR #104, opened 2026-06-25
Support booz_xform field input in addition to .bc

### LINKED ISSUE #103: Support booz_xform field input in addition to .bc
**Depends on:** itpplasma/libneo#344, itpplasma/libneo#345

## Goal

Read the Boozer field from a `booz_xform` `boozmn` NetCDF (Boozer Fourier harmonics), in addition to the current `.bc` input.

## Context

Shared converters live in libneo: EQDSK -> Boozer chartmap (itpplasma/libneo#343), `.bc` -> `booz_xform` (itpplasma/libneo#344), and a same-source converter/sign-convention gate (itpplasma/libneo#345). Prefer routing through libneo's existing chartmap/booz_xform reader over writing a new parser. Fortran codes can wrap the libneo reader via f90wrap or read the NetCDF directly; Python reads it directly.

Different Boozer transforms need not match to floating-point accuracy. This issue requires NEO-2 to import booz_xform coefficients with the correct local sign conventions.

## Files to edit

- `COMMON/neo_magfie.f90` (the Boozer harmonic field evaluator).

## What to add

- Accept a `boozmn` file as an alternative field source, mapping its harmonics into the existing `b_mn` Boozer representation.
- Cover booz_xform phase conventions explicitly: `cos(m theta - n zeta)` / `sin(m theta - n zeta)`, positive and negative `n`, and asymmetric coefficients such as `bmns_b`, `pmns_b`, and `pmnc_b`.

## Acceptance

- A reference run from a same-source converted `boozmn` input reproduces the `.bc` result within documented converter/interpolation tolerance. `.bc` remains supported.
- Tests fail for the known sign mistakes in booz_xform coefficient imports.

## Non-goals

- Removing `.bc` now. Changing the physics.
- Requiring agreement between independent Boozer transforms to floating-point accuracy.

krystophny:
Closes #103.

Adds Boozer field input from booz_xform (`boozmn`) and/or a Boozer chartmap in addition to the legacy `.bc`, reading via the libneo formats. Part of the `.bc` retirement effort (libneo converters merged in itpplasma/libneo#346).
Follow-up converter coverage for `.bc -> boozmn` sign handling is in itpplasma/libneo#357.

booz_xform stores Fourier modes as `m*theta - n*phi`. NEO-2's stellarator-symmetric path already evaluates that phase, so `ixn_b` is kept and `pmns_b` is imported as `lmnc = -pmns_b` to match the symmetric `phib - phi` convention. NEO-2's asymmetric path evaluates `m*theta + n*phi`, so asymmetric boozmn import flips `ixn = -ixn_b`, redirects through `INP_SWI_TOK`, and imports `lmnc = pmnc_b`, `lmns = pmns_b`.

The boozmn reader test covers axisymmetric `n=0`, stellarator-symmetric `n != 0`, asymmetric `n > 0`, asymmetric `n < 0`, `bmns_b`, `pmns_b`, and `pmnc_b`.

### Test fails before fix
Old code with the new phase/sign test fails on the dropped or wrong phase coefficients:

```text
$ cmake -S <old tree> -B build -G Ninja && cmake --build build --target test_boozmn_read -j$(nproc) && ctest --test-dir build -R boozmn_read_test --output-on-failure
FAIL: symmetric (m=1,ixn=+1) phase coefficient mismatch
FAIL: symmetric (m=2,ixn=-1) phase coefficient mismatch
FAIL: asymmetric (m=1,ixn=-1) phase coefficients mismatch
FAIL: asymmetric (m=2,ixn=+1) phase coefficients mismatch
FAIL
0% tests passed, 1 tests failed out of 1
```

### Test passes after fix
```text
$ fo && cmake --build build --target test_boozmn_read -j$(nproc) && ctest --test-dir build -R boozmn_read_test --output-on-failure && fo check
Test #19: boozmn_read_test .................   Passed    0.07 sec
Build: OK (0 modules, 0 cached, 0 changed, 0 affected) Tests: pass (.0s)
```


COMMON/neo_sub.f90
```diff
@@ -2570,4 +2578,332 @@ SUBROUTINE neo_write_bc
...
+      status = nf90_get_var(ncid, varid, lasym_int)
+    ELSE
+      lasym_int = 0
+    END IF
+
+    ! booz_xform stores Fourier coefficients for cos(m*theta - n*phi).
+    ! The asymmetric read path redirects through INP_SWI_TOK which evaluates
+    ! cos(m*theta + n*phi).  For n=0 (nboz_b=0) the sign is irrelevant.
+    ! For nboz_b>0 the mismatch is a correctness error; reject until fixed.
+    IF (lasym_int /= 0 .AND. nboz_b /= 0) THEN
```
GeorgGrassler: so if I get this correctly, the issue is that you do not know what the boozx_form convention is if we have the asymmetric case? Is this not documented in boozx_form? Cause I think https://hiddensymmetries.github.io/booz_xform/theory.html is pretty clear that the modes are to be understood as cos(m poloidal_angle - n toroidal_angle) etc for sin. You are in some sense always at odds here with the angle orientation, not just $n\neq 0$, cause the sin modes are effected by the theta orientation. Now the orientation of poloidal_angle_B and toroidal_angle_B, assuming the coordinate order is (radial, poloidal, toroidal) and radial always points outward, you can construct the orienation of the angles from the $(0,0) `gmnc_b`, $(1,0)$ `rmnc_b` and $(1,0)$ `zmns_b` modes. I am not saying that this fixes your problem, cause the other question is, what does `INP_SWI_TOK` actually expect. Just curious of the reason behind this guard.

COMMON/neo_sub.f90
```diff
@@ -2570,4 +2578,325 @@ SUBROUTINE neo_write_bc
...
+
+    status = nf90_inq_varid(ncid, 'ixn_b', varid)
+    IF (status /= nf90_noerr) THEN
+      WRITE(w_us,*) 'FATAL: boozmn missing variable ixn_b'
+      STOP
+    END IF
+    status = nf90_get_var(ncid, varid, ixn)
+    IF (lasym_int /= 0) THEN
+      ! INP_SWI_TOK evaluates asymmetric spectra with m*theta + n*phi.
+      ! Import booz_xform's m*theta - n*phi modes by reversing n at the boundary.
```
GeorgGrassler: remove "at the boundary" from the comment

COMMON/neo_sub.f90
```diff
@@ -2570,4 +2578,325 @@ SUBROUTINE neo_write_bc
...
+          zmnc(i,j) = zmnc_b(j,i)
+          zmns(i,j) = zmns_b(j,i)
+          lmnc(i,j) = pmnc_b(j,i)
+          lmns(i,j) = pmns_b(j,i)
+        END DO
+      END IF
+    END DO
+
+    ! Redirect asymmetric fields through INP_SWI_TOK evaluation path in neo_magfie.
+    IF (lasym_int /= 0) inp_swi = INP_SWI_TOK
```
GeorgGrassler, line 2848: interesting remark: historically #26 , there was already an explicit stellarator asymmetric case switch (I correct my earlier comments, I did use it in the past). So just to make sure, the `INP_SWI_TOK` only effects the spline constructions from modes? Cause if this is the case, we might be better off in renaming that to sth like `INP_SIW_ASYM` and the other `INP_SWI_SYM`. If there is more to it, leave it be, no need to make the pr more complicated.

COMMON/neo_sub.f90
```diff
@@ -2570,4 +2578,325 @@ SUBROUTINE neo_write_bc
...
+        ! INP_SWI_TOK evals z via zi*cosv + zis*sinv
+        ! where zi=s_zmnc (cos coeff) and zis=s_zmns (sin coeff).
+        DO j = 1, mnmax
+          bmnc(i,j) = bmnc_b(j,i)
+          bmns(i,j) = bmns_b(j,i)
+          rmnc(i,j) = rmnc_b(j,i)
+          rmns(i,j) = rmns_b(j,i)
+          zmnc(i,j) = zmnc_b(j,i)
+          zmns(i,j) = zmns_b(j,i)
+          lmnc(i,j) = pmnc_b(j,i)
```
GeorgGrassler, line 2841: about this, what is your reason for this non-flip? Cause in the end, the `pmns_b` follow the boozxform convention i.e. means sth very specific for the relation between $\varphi$ and $\varphi_B$. When you "missuse" the `INP_SWI_TOK` to funnel in asymmetric fields, you should make sure that the metric that you now imply with this is understood downstream the same way.

**If** you did this non-flip, because you just saw the previous `INP_SWI_TOK` branch not do the flip when reading from the `.bc`, be aware this could be just because the respective `.bc` files were written that way i.e. they included the extra sign already. **Only if** you did not already check that, a way to confirm is to sanity check the computed $< \nabla s>$ and geometrical surface.

You can use for example the

```python
from neo2_ql import get_fluxsurface_area
```

to compute the area from the NEO-2 output and compare that to the area suggested from the original VMEC .nc file (I assume there is a reader/plotter for that in simsopt e.g. `from simsopt.geo import SurfaceRZFourier`). The difference should be quite stark (test by running with the opposite sign) as the surface gets quite distorted if the wrong convention is used, see also 

https://gitlab.tugraz.at/plasma/proj/convert_qps_boozer_file/-/blob/main/graphs/no_conversion.png?ref_type=heads

vs

https://gitlab.tugraz.at/plasma/proj/convert_qps_boozer_file/-/blob/main/graphs/v_function_conversion_3d_boundary.png?ref_type=heads

Though I am unsure how this could be included as a proper test beyond an initial sanity check as this pipeline needs a full NEO-2 run currently.

COMMON/neo_sub.f90
krystophny: Done in f247ffb: dropped " at the boundary", the comment now reads `! Import booz_xform's m*theta - n*phi modes by reversing n.`

COMMON/neo_sub.f90
krystophny: It affects more than spline construction, so I would keep the name. `INP_SWI_TOK` (=9, historically ASDEX-U / E. Strumberger) is the numeric `inp_swi` mode read from the input file, and it gates the whole asymmetric evaluation path, not just spline setup:

- `neo_magfie.f90`: allocation and splining of the sin arrays (`s_bmns`, `s_rmns`, `s_zmns`, `s_lmns` and their radial derivatives), the Fourier summation convention itself (`cosv/sinv` built as `m*theta + n*phi`, the `+...*sinv` asymmetric terms, the `b_tb`/`b_pb` signs), the toroidal-angle conversion sign (`p = x(2) + l*twopi/nfp` vs `- ...`), and the `G_symm` sign.
- `NEO-2-QL/neo_magfie_perturbation.f90`: the same switch drives the perturbation-field branches.
- `neo_sub.f90`: input reading/labeling for the Strumberger tokamak files.

So it is really "asymmetric-tokamak evaluation convention," and the boozmn path deliberately reuses it (`inp_swi = INP_SWI_TOK`) to land on that evaluator. Renaming to `INP_SWI_ASYM`/`INP_SWI_SYM` would mislead on two counts: the symmetric counterpart is not one "SYM" mode (`INP_SWI_STEL`, `INP_SWI_STEL_QPS`, and the `lab_swi` CHS/LHD branch all share the else path), and `INP_SWI_TOK` is a fixed numeric contract in existing input files. Per your "if there is more to it, leave it be" - there is, so I will keep the name and, if useful, add a doc comment noting it selects the asymmetric evaluation convention.

COMMON/neo_sub.f90
krystophny: The sign convention now lives in libneo, not here. After #357 the boozmn file carries `pmnc_b`/`pmns_b` and `ixn_b = n*nfp` in booz_xform's `cos/sin(m*theta - n*phi)` convention, checked in that PR against booz_xform's own `ixn_b` signs and against a public asymmetric Simsopt VMEC file. On the NEO-2 side the only convention change happens at import (`neo_read_boozmn`, commit "Convert asymmetric boozmn n sign on import"): `ixn = -ixn`, so `m*theta - n*phi` becomes INP_SWI_TOK's `m*theta + n*phi` on the identical numerical basis. The coefficients keep their booz_xform values; only the angle index is reversed.

Given that, the `pmns_b` non-flip is required, not carried over blindly from the old branch. The two lambda paths in `neo_magfie` convert phi with opposite signs:

```
IF (inp_swi == INP_SWI_TOK) THEN
   p = x(2) + l * twopi / nfp
ELSE
   p = x(2) - l * twopi / nfp
END IF
```

(the `+l` is the 2015 Strumberger correction). To make both paths yield the same physical mapping `phi - phi_b = pmns_b * sin(...)`, the symmetric path stores `lmnc = -pmns_b` while the asymmetric/TOK path stores `lmns = +pmns_b, lmnc = +pmnc_b`. Flipping in the asymmetric path would combine with the `+l` evaluator and invert `phi - phi_b`. So the sign difference between the two branches is exactly the downstream metric being kept consistent, which is your concern. Pass 4 of `test_boozmn_read` pins this (`lmns = +pmns_b`, `lmnc = +pmnc_b`).

That is reader/evaluator consistency, though, not an end-to-end geometry check. The `get_fluxsurface_area` vs `SurfaceRZFourier` cross-check has not been run and is pending. Note it mainly exercises R,Z plus the `ixn` sign (surface shape), i.e. the `ixn = -ixn` conversion; the `pmns` direction only shows up in the metric-dependent quantities (`p_pb`, `<|grad s|>`). I will run both: rebuild the surface with the opposite `ixn` sign to confirm the undistorted area against the VMEC `.nc`, and separately check a `<|grad s|>`-type quantity for the `pmns` sign. Agreed it needs a full NEO-2 run, so it stays a sanity check rather than a unit test for now.

GeorgGrassler: > ```
> IF (inp_swi == INP_SWI_TOK) THEN
>    p = x(2) + l * twopi / nfp
> ELSE
>    p = x(2) - l * twopi / nfp
> END IF
> ```
> 
> (the `+l` is the 2015 Strumberger correction). To make both paths yield the same physical mapping `phi - phi_b = pmns_b * sin(...)`, the symmetric path stores `lmnc = -pmns_b` while the asymmetric/TOK path stores `lmns = +pmns_b, lmnc = +pmnc_b`. Flipping in the asymmetric path would combine with the `+l` evaluator and invert `phi - phi_b`. So the sign difference between the two branches is exactly the downstream metric being kept consistent, which is your concern. Pass 4 of `test_boozmn_read` pins this (`lmns = +pmns_b`, `lmnc = +pmnc_b`).
> 

Thanks that is sufficient, just wanted to make sure that you keep that into account.

COMMON/neo_sub.f90
GeorgGrassler: fine by me. But be aware that we are also not honest here, cause `INP_SWI_TOK` would suggest that the field is axisymmetric here, which it clearly is not. And `neo_magfie_pertubation.f90`, this does not really apply for our toolchain cause you read all from a single boozxform file i.e. the field in your case is one thing, not axisymetric + pertubation. So we are misusing the tokamak workflow for sth that is actually stellarator in honesty, as seen by our flipping of `pmnc` and `n` conventions. Not saying you should fix this inconsitency now, just saying the real solution would be to decouple asymmetric/symmetric from tokamak/stellarator conventions (like the sign). Should be a future issue to be resolved.

GeorgGrassler:
@krystophny I added my comment to the `INP_SWI_*` aspect. TD;LR: make the remaining inconsistency its own issue and merge the current pr.

COMMON/neo_sub.f90
krystophny: Fixed in f247ffb.

COMMON/neo_sub.f90
krystophny: Filed as #115: decouple symmetric/asymmetric spectra from the tokamak/stellarator conventions so asymmetric boozmn input no longer masquerades as INP_SWI_TOK.

krystophny:
@GeorgGrassler filed the convention decoupling as #115 and the shared-reader follow-up as #116 (libneo now provides the raw boozmn reader, itpplasma/libneo#364). Merging as agreed.

PR #105, opened 2026-06-29
Assemble spline system directly as sparse

krystophny:
## Summary
- assemble the `splinecof3` linear system directly into sparse row/column/value arrays
- remove the dense `MA(:,:)` allocation and dense-to-sparse conversion from the spline solve path
- add a focused spline reconstruction test that exercises the sparse assembly path

### Test fails on main
```text
$ git show origin/main:COMMON/spline_cof.f90 | rg -n "CALL sparse_solve\(MA, inh\)|ALLOCATE\(MA\("
170:  ALLOCATE(MA(size_dimension, size_dimension),  stat = i_alloc, errmsg=error_message)
564:  CALL sparse_solve(MA, inh)
```

### Test passes after fix
```text
$ ctest --test-dir build -R spline_sparse_assembly_test --output-on-failure
Test project /tmp/neo2-spline-sparse-assembly/build
    Start 21: spline_sparse_assembly_test
1/1 Test #21: spline_sparse_assembly_test ......   Passed    0.06 sec

Total Test time (real) =   0.06 sec
```

```text
$ fo
# exited 0 with no output
```

### DIFF
--- a/TEST/CMakeLists.txt
+++ b/TEST/CMakeLists.txt
@@ -32,3 +32,16 @@ set_tests_properties(er_rotation_test PROPERTIES
     FAIL_REGULAR_EXPRESSION "FAIL"
     TIMEOUT 30
 )
+
+add_executable(test_spline_sparse_assembly test_spline_sparse_assembly.f90)
+target_link_libraries(test_spline_sparse_assembly common)
+
+add_test(NAME spline_sparse_assembly_test
+         COMMAND test_spline_sparse_assembly
+         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+
+set_tests_properties(spline_sparse_assembly_test PROPERTIES
+    PASS_REGULAR_EXPRESSION "All tests passed!"
+    FAIL_REGULAR_EXPRESSION "FAIL"
+    TIMEOUT 30
+)
--- a/TEST/CMakeLists.txt
+++ b/TEST/CMakeLists.txt
@@ -42,6 +42,5 @@ add_test(NAME spline_sparse_assembly_test
 
 set_tests_properties(spline_sparse_assembly_test PROPERTIES
     PASS_REGULAR_EXPRESSION "All tests passed!"
-    FAIL_REGULAR_EXPRESSION "FAIL"
     TIMEOUT 30
 )
--- a/TEST/test_spline_sparse_assembly.f90
+++ b/TEST/test_spline_sparse_assembly.f90
@@ -1,0 +1,52 @@
+program test_spline_sparse_assembly
+  use inter_interfaces, only: splinecof3, splint_horner3
+  use nrtype, only: DP, I4B
+
+  implicit none
+
+  integer(I4B), parameter :: n = 8
+  real(DP) :: x(n), y(n), lambda(n)
+  real(DP) :: a(n), b(n), c(n), d(n)
+  integer(I4B) :: indx(n)
+  real(DP) :: c1, cn, y_eval, yp, ypp, yppp
+  real(DP) :: max_err
+  integer(I4B) :: i
+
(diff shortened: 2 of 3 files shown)


TEST/CMakeLists.txt
```diff
@@ -32,3 +32,16 @@ set_tests_properties(er_rotation_test PROPERTIES
...
+
+add_executable(test_spline_sparse_assembly test_spline_sparse_assembly.f90)
+target_link_libraries(test_spline_sparse_assembly common)
+
+add_test(NAME spline_sparse_assembly_test
+         COMMAND test_spline_sparse_assembly
+         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+
+set_tests_properties(spline_sparse_assembly_test PROPERTIES
+    PASS_REGULAR_EXPRESSION "All tests passed!"
```
GeorgGrassler, line 44: Instead of the fragile regular expression matching, consider make the test error stop on failure. This already notes the test as a failure. You can keep the timout setting though

GeorgGrassler: I am aware that other tests have the same pattern, but I would argue it is not a good pattern

GeorgGrassler: nevermind I checked, `stop 1` is already throwing errors, so the regex is optional ... can be left or removed

TEST/test_spline_sparse_assembly.f90
```diff
@@ -0,0 +1,52 @@
...
+       a, b, c, d, 1.0_DP, unit_weight)
+
+  max_err = 0.0_DP
+  do i = 1, n
+    call splint_horner3(x, a, b, c, d, 0_I4B, 1.0_DP, x(i), &
+         unit_weight, zero_weight, zero_weight, zero_weight, y_eval, yp, ypp, yppp)
+    max_err = max(max_err, abs(y_eval - y(i)))
+  end do
+
+  if (max_err > 1.0e-10_DP) then
```
GeorgGrassler: @krystophny question: those are smoothed splines with `lambda` the smoothing parameter? I assume that is the reason that we do not go exactly through the spline points i.e. 1e-10? Would we get exact interpolating spilnes if `lambda=0`? If yes, worth to consider to also include this test case. If not, and this is more complicated, the current state is fine.

TEST/CMakeLists.txt
krystophny: Pushed 060ff82: dropped `FAIL_REGULAR_EXPRESSION` since `stop 1` already fails ctest via exit code regardless of the regex. Kept `PASS_REGULAR_EXPRESSION` though — I checked, it's not pure redundancy: it still catches a test that exits 0 without ever reaching the final print (e.g. a silently skipped assertion), which a bare exit-code check would miss. Scoped the change to this new test block only, left the existing pattern elsewhere untouched.

TEST/test_spline_sparse_assembly.f90
krystophny: Dug into `splinecof3` (`COMMON/spline_cof.f90:219-225`): the `lambda` argument we pass gets remapped internally — `omega = lambda1`, and the actual smoothing penalty added to the matrix diagonal is `1 - omega`. So `lambda=1.0` here is already the exact-interpolation case (full data weight, zero smoothing), not a smoothed fit — naming is inverted from what it suggests.

Verified numerically with this dataset: `lambda=1.0` reconstructs to ~1.1e-13 (floating-point round-off, stable across Debug and Release builds, no jitter across repeated runs), while `lambda=0.0` is degenerate — zero data weight, pure curvature minimization, singular for this system (NaN). So there's no `lambda=0` exact-interpolation case to add; the existing test already covers exact interpolation, just with a looser-than-necessary tolerance.

Pushed 060ff82 tightening `1.0e-10_DP` to `1.0e-12_DP` to actually assert exactness (~9x margin over the observed error).

GeorgGrassler (approved):

PR #106, opened 2026-06-29
Investigate banded spline solve path

krystophny:
Depends on #105 (merged). Keeps the generic sparse path for non-common switch
combinations and adds a direct LAPACK band solve for the common `splinecof3`
case `sw1=2, sw2=4`, which is the only combination any physics caller uses.

## What changes

- `sw1=2, sw2=4` assembles directly into LAPACK band storage and solves with
  `dgbsv` after a fixed symmetric unknown/equation reordering
  (`init_common_band_order`, block variable order `[5,6,7,1,2,3,4]`, `VAR=7`).
  Band profile `lower_band=17, upper_band=14` on the recorded golden matrices.
- All other switch combinations continue through the sparse assembly and
  SuiteSparse (UMFPACK) solve from #105.
- The common path has no sparse fallback and drops the earlier Woodbury split,
  whose chosen band core was singular in CI.

Trace basis: the extracted `splinecof3` records contain 608 cases, all
`sw1=2, sw2=4`. Physics callers that now take the band path (all `sw1=2, sw2=4`):

- `COMMON/neo_sub.f90`: iota, pprime, sqrtg00, curr_tor, curr_pol, B00 radial
  profiles.
- `COMMON/collop_spline.f90`: collision-operator matrix elements.

## Golden-record deviations (manual local check)

Built this branch and `main`, ran the NEO-2 golden record (`lorentz`, `ql`) with
each build as the other's reference.

Root cause of every deviation: the common case now factorizes with LAPACK
`dgbsv` (banded LU, partial pivoting) instead of UMFPACK (sparse multifrontal
LU). The different pivot order produces different round-off in the spline
coefficients, ~1e-10 relative, matching the 608-record trace. Derivative and
geometry quantities that read those coefficients directly carry the largest
relative change; integrated transport averages it down.

Lorentz, this branch vs `main`:

- Magnetics field-derivative and geometry: `d2bp_ext/min/max` up to 5.2e-11,
  `dlogbdphi` 3.9e-11, `geodcu` 2.7e-11, `x3` 1.4e-11.
- Transport: `D11_ov_Dpl` 7.1e-14, `dmono_over_dplateau` 7.1e-14,
  `epseff3_2` 5.7e-14, `g_bs` 3.4e-14, `qflux`/`qcurr` 1-2e-14,
  `aiota_loc` 1.4e-14.

QL, this branch vs `main`:

- Largest: `D11_AX` 7.6e-11, `dbcovar_phi_ds` 6.6e-11, `D21_AX` 6.2e-11,
  `D12_AX` 5.0e-11, `TphiNA_Ware_tot` 4.1e-11, `Qflux_NA_Ware_spec` 4.0e-11,
  `Gamma_NA_Ware_spec` 2.9e-11, `D23_NA`/`D13_NA` ~2.5e-11.
- Most spec, flux and torque outputs: 1e-12 to 1e-13.
- Geometry (`aiota`, `bcovar_phi`, `bcovar_tht`, `sqrtg_bctrvr_*`, `avbhat`):
  1e-14 to 1e-15.

Maximum relative deviation across both cases: 7.6e-11. These are round-off
differences from the LU-backend swap, not a physics change.

## Effect on the golden gate

- Frozen committed golden (`legacy`, in the data repo): lorentz passes at
  `LEGACY_TOL=1e-10`, ql at `LEGACY_TOL=1e-8`, par at `LEGACY_TOL=2e-8`. The
  drift stays under these bars, so no committed golden values need
  regenerating.
- Exact main-reference check (`run_test_main`, reference built from `main`):
  once #105 and #106 are on `main`, that reference already carries this backend,
  so current equals reference and the tight tolerances (lorentz
  `TOL=1e-14`/`MAGNETICS_TOL=1e-12`, ql `TOL=1e-15`) pass. Kept tight. Before
  both land, this leg is expected red because `main` is still UMFPACK.
- Rough stable check (`run_test_stable`, reference built from the latest
  release, still UMFPACK): the only leg that spans the backend change. Widened
  in `test-on-pr.yml` to the measured drift plus margin (lorentz `TOL=1e-12`,
  `MAGNETICS_TOL=1e-9`; ql `TOL=1e-9`). Once a release cut from this code
  becomes the latest, the drift disappears and these can be tightened again.

Tolerance changes live in the workflow files, not the data runfolders.

## Verification

Merge of `main` is clean. Only `spline_cof.f90` and the spline test conflicted:
`main`'s #105 squash is byte-identical to this branch's ancestor, and the test
here is a strict superset that also asserts the `sw1=2, sw2=4` reconstruction.

Unit test on the merged branch:

```text
$ ctest -R spline --output-on-failure
1/1 Test #3: spline_sparse_assembly_test ......   Passed    0.06 sec
```

Golden record, exact leg fails against an UMFPACK reference (numbers above),
frozen-golden leg passes:

```text
$ python compare_h5.py reference/summary.h5 current/summary.h5 1e-15   # ql
ValueError: current and reference are not the same

$ python compare_h5.py legacy/summary.h5 current/summary.h5 1e-8       # ql
(no differences above tolerance)

$ python compare_h5.py legacy/summary.h5 current/summary.h5 1e-10      # lorentz
(no differences above tolerance)
```

### DIFF
--- a/.github/workflows/test-on-pr.yml
+++ b/.github/workflows/test-on-pr.yml
@@ -103,6 +103,8 @@ jobs:
 
       - name: Build NEO-2 (reference version - latest stable release)
         id: build_reference_stable
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           # Build reference version in temp directory
           REFERENCE_DIR=$(mktemp -d)
@@ -110,8 +112,12 @@ jobs:
 
           git clone https://github.com/itpplasma/NEO-2.git .
 
-          # Get latest stable release tag from GitHub API
-          LATEST_RELEASE=$(curl -s https://api.github.com/repos/itpplasma/NEO-2/releases/latest | jq -r '.tag_name')
+          # Get latest stable release tag from GitHub API. Authenticated with
+          # the run's own GH_TOKEN: an anonymous curl call here shares the
+          # unauthenticated 60 req/hour rate limit across every concurrent CI
+          # run in the repo and starts failing under load.
+          LATEST_RELEASE=$(curl -s -H "Authorization: Bearer ${GH_TOKEN}" \
+            https://api.github.com/repos/itpplasma/NEO-2/releases/latest | jq -r '.tag_name')
           if [ -z "$LATEST_RELEASE" ] || [ "$LATEST_RELEASE" == "null" ]; then
             echo "Error: Failed to fetch the latest release tag from GitHub API." >&2
             exit 1
--- a/.github/workflows/test-on-pr.yml
+++ b/.github/workflows/test-on-pr.yml
@@ -166,12 +166,16 @@ jobs:
           cd "$TEST_DIR"
           echo "test_dir=$(pwd)" >> $GITHUB_OUTPUT
 
-          export TOL=1e-14
+          # Rough check against the pre-106 release: the sparse->band splinecof3
+          # backend swap shifts spline-derivative quantities by up to ~5e-11
+          # (magnetics) and integrated transport by ~1e-13. Widen only this
+          # stable comparison; the exact main-reference check stays tight.
+          export TOL=1e-12
           export LEGACY_TOL=1e-10
-          export MAGNETICS_TOL=1e-12
+          export MAGNETICS_TOL=1e-9
           make lorentz
 
-          export TOL=1e-10
+          export TOL=1e-9
           export LEGACY_TOL=1e-8
           make ql
 
--- a/.github/workflows/unit-tests-coverage.yml
+++ b/.github/workflows/unit-tests-coverage.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Set up Python 3
         run: |
           python -m pip install --upgrade pip
-          pip install numpy h5py scipy
+          pip install numpy h5py scipy matplotlib
 
 
 
--- a/.github/workflows/unit-tests-coverage.yml
+++ b/.github/workflows/unit-tests-coverage.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Set up Python 3
         run: |
           python -m pip install --upgrade pip
-          pip install numpy
+          pip install numpy h5py scipy
 
 
 
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Set up Python 3
         run: |
           python -m pip install --upgrade pip
-          pip install numpy h5py scipy
+          pip install numpy h5py scipy matplotlib
 
       - name: Build NEO-2
         run: |
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Set up Python 3
         run: |
           python -m pip install --upgrade pip
-          pip install numpy
+          pip install numpy h5py scipy
 
       - name: Build NEO-2
         run: |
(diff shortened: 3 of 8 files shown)


GeorgGrassler (approved):
regarding the results of the golden record test:

The spikes in numerical drifts look plausible e.g. 

```
Difference in dbcovar_phi_ds: 1.510126757025429e-08
Difference in dbcovar_theta_ds: 2.3981486096758673e-11
```

where the actual value of $dB_\varphi/ds$ is 3-4 orders smaller than $dB_\vartheta/ds$ looking at `performance_ql/test_axi.bc`. On this heuristic sanity check drifts seem to originate purely from machine arithmetic differences. Additional note: the higher errors at the performance test vs the normal ql benchmark might be due to a factor 200 denser grid in s (compare the `test_axi.bc` for the two benchmarks). You might let Claude reason itself through that if you like.

For a sparse s grid benchmark (e.g. the check agains the latest stable version) the errors drift are fully exceptable like the sensitive "magnetics" quantities alias `d2bp_ext` etc.

@krystophny From my side those numbers are fine and the pr can be merged once #105 lands. 

## regarding tolerances

For all benchmarks except the first one, the reference version is `main`. Therefore, you do not need to widen the tolerances, cause this will become the new reference and subsequent prs will pass anyway. If you change again sth in the fundamental numerics, this golden record will let you know that you changed sth ... which is the intention i.e. I want a change in the numerics to be a conscious choice even if it signals that by noise as explained above (i.e. the gates block auto-merges). 

For the stable version comparison you can **widen the tolerances a bit**, the violations there where a mild drift anyway. That way, when you do more changes on the spline side (i.e. numerics noise appears again) and both the stable version benchmark and the rough check of `par` pass and only the tight ones fail, you know that you changed sth, but not physically relevant i.e. you can bypass the gates/merge by hand.

krystophny:
Thanks for the sanity check on the drift and the tolerance guidance — makes sense, will leave the main-reference tolerances tight and only widen the stable-version comparison a bit. Also pushed the #105 tolerance/regex fixes here via cherry-pick (a4dccba). Will merge both #105 and #106 and update the golden record reference values, unless you have further objections.

GeorgGrassler:
> Thanks for the sanity check on the drift and the tolerance guidance — makes sense, will leave the main-reference tolerances tight and only widen the stable-version comparison a bit. Also pushed the #105 tolerance/regex fixes here via cherry-pick ([a4dccba](https://github.com/itpplasma/NEO-2/commit/a4dccbab9233155d9e5eff51c5725ca82d4a9e71)). Will merge both #105 and #106 and update the golden record reference values, unless you have further objections.

just to be sure: updating golden record reference values means here updating main, or are you thinking of updating the golden record file in case of the `par` run? Cause this one already passes and does not look like it needs updating. Also, the updating of the tolerances should happen in the `yml` workflow files, not anything in the backend `$DATA´ runfolders.

krystophny:
Right on both counts.

Reference values don't change. Against the frozen `legacy` oracle in `$DATA`
this branch stays within tolerance (lorentz 1e-10, ql 1e-8, par 2e-8), so no
golden files get rewritten -- par included, which is why it already passes.
"Updating reference values" just means `run_test_main` builds its reference from
`main`: once #105 and #106 are both on `main`, current == reference and the
exact check passes at the tight tolerances.

The only leg spanning the backend change is `run_test_stable` (reference =
pre-106 release, still UMFPACK). Widened just that one in `test-on-pr.yml`, the
main-reference tolerances left tight: lorentz TOL 1e-14->1e-12, MAGNETICS_TOL
1e-12->1e-9; ql TOL 1e-10->1e-9. Nothing in `$DATA`.

Sizing from a local run of this branch vs `main`: max relative drift 7.6e-11, in
spline-derivative/geometry quantities (lorentz d2bp ~5.2e-11, dlogbdphi 3.9e-11;
ql D11_AX 7.6e-11, dbcovar_phi_ds 6.6e-11); integrated transport ~1e-13-1e-14.
Root cause is the sw1=2,sw2=4 splinecof3 solve moving from UMFPACK to LAPACK
dgbsv (different pivot order).

After merge, a release cut from the merged commit becomes the new "latest" and
the stable tolerances can go back tight.

Issue #118, opened 2026-07-06
QL: single-helicity misalignment drive (E×B + v_par dB^s) in ripple_solver with aligned-pair zero-current regression


krystophny:
**Stage:** NEO-2-QL, non-axisymmetric (quasilinear) solve in `ripple_solver_ArnoldiO2`
**Source language:** Fortran
**Manuals to read first:** `DOC/ripple_solver_normalizations_and_output.tex` (solution vector, flux matrix, dimensional coefficients); Kasilov et al., PoP 21, 092506 (2014); Markl et al., NF 63, 126007 (2023) Sec. 2 (thermodynamic forces $A_1$, $A_2$ and the perpendicular misalignment field)

## Goal

Add an optional additional right-hand side to the non-axisymmetric solve of `ripple_solver_ArnoldiO2`: the drive of a single-helicity perturbation given in the unshifted frame by a radial magnetic corrugation $\delta B^s$ together with an electrostatic potential harmonic $\Phi_{mn}$,

$$S = -\left(v_\parallel\,\frac{\delta B^s}{B_0} + \tilde v_E^s\right)\frac{\partial f_0}{\partial s}, \qquad \tilde v_E^s = \frac{c\,(\mathbf B_0\times\nabla\delta\Phi)\cdot\nabla s}{B_0^2},$$

with $\partial f_0/\partial s \to f_0\,(A_1 + x^2 A_2)$, $x = v/v_{T}$, decomposed over the standard force channels so that the existing transport-matrix machinery (`qflux`, forces applied downstream) is reused unchanged. This makes the quasilinear parallel-current response to a *misaligned* electrostatic perturbation (equipotentials not aligned with perturbed flux surfaces) computable with NEO-2-QL, which is the kinetic drive relevant for RMP shielding currents and helical-core (flux-pumping) current redistribution. The default behavior (switch off) must be bit-identical to current `main`.

Both perturbation pieces are single Fourier harmonics $\propto \exp(im\vartheta + in\varphi)$ in Boozer angles, i.e. $\propto \exp\big(i(m\iota + n)\varphi\big)$ along the field line — same phase convention as the existing `bnoverb0_test` scaffold (`ripple_solver_ArnoldiOrder2_test.f90:942-947`).

## Physics to implement

Amplitudes are flux functions times the helical phase. With $c_B \equiv (\delta B^s/B_0^\varphi)_{mn}$ (dimensionless, complex) and $\Phi_{mn}$ (statvolt, complex):

[...]

PR #119, opened 2026-07-06
QL: single-helicity misalignment drive in ripple_solver_ArnoldiO2

krystophny:
Implements #118. This branch is stacked on #124 and #121 and remains open against `main`. The separate child branch for #122 enables source-only execution when this drive supplies a nonzero source.

## Summary

- add `isw_hel_drive`, `m_theta_hel`, `hel_brad_re/im`, and `hel_phim_re/im` to `ntv_input`
- add the single-helicity right-hand side to the non-axisymmetric Arnoldi O2 solve, default off
- add the two velocity moments needed by the drive and unload their storage with the other collision moments
- expose and test the production source-activation predicate used by the #122 child
- document input phase, amplitudes, and CGS units

## Invariants

Switch-off behavior remains unchanged. The drive preserves the existing Arnoldi O2 discretization, collision-moment storage layout, species ordering, ABI, field-line boundary conditions, and Fourier convention. The source uses CGS/Gaussian units and the CAS-checked `exp(+i(m theta + n phi))` phase map. No branch in this stack is authorized to merge before the programme-wide gate.

### Test fails on main

With the source-activation regression applied before the drive implementation:

```text
$ cmake --build --preset default --target test_ntv_hel_drive_input
Error: Symbol ‘has_helical_drive_source’ referenced at (1) not found in module ‘ntv_mod’
Error: Function ‘has_helical_drive_source’ at (1) has no IMPLICIT type
ninja: build stopped: subcommand failed.
```

### Test passes after fix

```text
$ fo
Static: OK
Build: OK
Tests: OK (9/9)
Lint: OK
All stages passed
```

The full pipeline passed twice consecutively after #124 serialized the duplicate build tests.

```text
$ make test
100% tests passed, 0 tests failed out of 7

$ ctest --test-dir build -R "^ntv_hel_drive_input_test$" --output-on-failure
1/1 Test #...: ntv_hel_drive_input_test ... Passed

$ git diff --check
<no output>
```

The 16 separately managed external-data regressions remain registered under `regression;external-data`; this branch does not mark that unavailable-data gate passed.

PR #121, opened 2026-07-07
QL: allow prescribed perturbation mode without field file

krystophny:
This branch is stacked on #124 and remains open against `main`. PR #120 is already merged.

## Summary

- add `isw_m_phi_input` and `m_phi_input` to `ntv_input`
- allow file-less non-axisymmetric input with a prescribed toroidal mode
- reject a file-less solve when no physical source is configured
- skip perturbation-file symlinks when no file is requested
- document and test the input path

The #118 drive is the next stack layer. Its #122 child relaxes the no-source guard only when a nonzero helical source exists. No branch is authorized to merge before the programme-wide gate.

### Test fails on main

With only the regression applied:

```text
$ cmake --build --preset default --target test_ntv_perturbation_input
Error: Symbol ‘isw_m_phi_input’ referenced at (1) not found in module ‘ntv_mod’
Error: Symbol ‘m_phi_input’ referenced at (1) not found in module ‘ntv_mod’
Error: Symbol ‘has_perturbation_file’ referenced at (1) not found in module ‘ntv_mod’
ninja: build stopped: subcommand failed.
```

### Test passes after fix

```text
$ fo
Static: OK
Build: OK
Tests: OK
Lint: OK
All stages passed

$ make test
100% tests passed, 0 tests failed out of 6

$ ctest --test-dir build -R "^ntv_perturbation_input_test$" --output-on-failure
1/1 Test #...: ntv_perturbation_input_test ... Passed

$ git diff --check
<no output>
```

All 16 external-data regressions remain registered under `regression;external-data`; this branch does not mark that separate gate passed.

### DIFF
--- a/DOC/neo2.in.ql-full
+++ b/DOC/neo2.in.ql-full
@@ -167,9 +167,12 @@
 &ntv_input
  B_rho_L_loc = 0.0 ! Larmor radius times magnetic field. [0.0d0]
  in_file_pert = 'filename' ! File that contains perturbation of the
-                           ! magnetic field.
+                           ! magnetic field. Use 'none' for source-only
+                           ! non-axisymmetric runs with prescribed m_phi.
  isw_mag_shear = 1 ! Turn on(=1)/off(=0) non-local computations (mag.
                    ! shear). [0]
+ isw_m_phi_input = 0 ! Use m_phi_input instead of the perturbation-file mode.
+                     ! Required when in_file_pert='none'. [0]
  isw_ntv_mode = 1  ! [0]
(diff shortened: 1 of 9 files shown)


krystophny:
The `in_file_pert='none'` source-only path this PR adds lets an NA qflux run (`isw_qflux_NA=1`) proceed with no source configured (no field, drive off). The resulting zero source is the Arnoldi seed, so `arnoldi` normalizes 0/0 -> NaN and `zhseqr` fails at every step (~8 min spin, no qflux output). Scoped guard + red-then-green test: #122. Keep `arnoldi_mod.f90` untouched; guard the assembled source (a valid `none` + drive-on run on #119 must still pass).

krystophny:

PR #124, opened 2026-07-11
Classify external-data CTests for local pipeline

krystophny:
## Summary

The CMake tree registers 16 reconstruction tests that require separately managed templates and reference datasets. The supported local `make test` target deliberately runs the self-contained tests under `build/TEST`, but generic CTest discovery sees both groups.

This change labels the reconstruction cases `regression;external-data`, keeping all 16 registered and runnable with the complete external-data workflow while allowing the default local pipeline to exercise the self-contained suite. It also removes two unused imports reported by the full lint pass.

No numerical algorithm, discretization, convergence criterion, source term, generated code, ABI, memory layout, physical convention, boundary condition, or conservation law changes. Test commands and expected numerical comparisons are unchanged.

### Test fails on main

```text
$ NEO2_TEST_PATH=<existing-temporary-directory> fo
Static: OK
Build: OK
Tests: FAIL
9 passed, 16 failed
```

All 16 failures stop in harness setup because the separately managed templates and references are unavailable; no numerical comparison is reached. The subsequent full lint pass also reports the two unused imports.

### Test passes after fix

```text
$ fo
Static: OK (129 modules)
Build: OK
Tests: skipped, no affected tests
Lint: OK
All stages passed

$ make test
100% tests passed, 0 tests failed out of 5

$ ctest --test-dir build -N -L external-data
Total Tests: 16

$ git diff --check
<no output>
```

The 16 numerical regressions remain a separate unverified gate until their external datasets are supplied; this PR does not mark them passed.


krystophny:

PR #125, opened 2026-07-11
QL: guard source-only non-axisymmetric execution

krystophny:
Closes #122. This child is stacked on #119, #121, and #124 and remains open against `main`.

## Summary

- reject a fileless non-axisymmetric solve when neither a perturbation file nor an active, nonzero helical drive supplies the source
- use the production helical-source predicate shared with the drive implementation
- cover rejected zero-source and accepted driven-source configurations through the public initialization and validation path

## Invariants

The guard runs before Arnoldi assembly. It does not change the numerical algorithm, discretization, convergence criteria, generated code, ABI, storage layout, Fourier convention, CGS units, source signs, boundary conditions, or switch-off response. A valid nonzero helical source passes unchanged; an identically zero source terminates before zero-norm normalization can create NaNs.

No branch in this stack is authorized to merge before the programme-wide gate.

### Driven case fails on parent

The two process-level tests were applied without the production change at parent `7a868ae`:

```text
$ ctest --test-dir build -R '^ntv_(zero|driven)_source_guard_test$' --output-on-failure
50% tests passed, 1 tests failed out of 2
ntv_zero_source_guard_test   Passed
ntv_driven_source_guard_test Failed
ERROR: non-axisymmetric solve has no source.
```

The parent incorrectly rejected `isw_hel_drive=1` with nonzero `hel_phim_re`.

### Both cases pass after fix

```text
$ fo
Static: OK
Build: OK
Tests: OK
Lint: OK
All stages passed

$ make test
100% tests passed, 0 tests failed out of 9

$ ctest --test-dir build -R '^ntv_(zero|driven)_source_guard_test$' --output-on-failure
100% tests passed, 0 tests failed out of 2
```

The zero-source test emitted the single expected diagnostic and returned in under five seconds. The driven test emitted `Driven source accepted!`. Neither output contained `NaN` or `zhseqr`.

```text
$ git diff --check
<no output>
```

The 16 separately managed external-data regressions remain registered under `regression;external-data`; this branch does not mark that unavailable-data gate passed.

PR #126, opened 2026-07-11
QL: preserve full-step endpoints after mode-3 refinement

### LINKED ISSUE #123: QL: source_flux reads npl(istep+-2) past bounds -> heap corruption on helical-axis geometries
## Summary

A QL non-axisymmetric solve (`isw_qflux_NA=1`) on a helical-axis geometry that triggers the `phi_divide>1` field-line re-split aborts. Under `-fcheck=all` it stops with an `npl` index over-read in `source_flux`; in a `Release` build the over-read produces a garbage neighbour count that then drives an out-of-bounds write, and the process dies later with glibc `corrupted size vs. prev_size` at the post-solve `deallocate`.

## Reproduction

- QL run, `isw_qflux_NA=1`, on a helical-axis Boozer field whose field-line splitting yields an odd `(iend - ibeg)` and hits `maxval(phi_divide) > 1` (so the solver re-splits with `phi_split_mode=3`, `max_solver_try>1`).
- Build with bounds checks: `cmake -DCMAKE_Fortran_FLAGS="-fcheck=all -fbacktrace -g -O0" . && ninja neo_2_ql.x`, then run the case.

Observed (`-fcheck=all`):

```
At line 4135 of file NEO-2-QL/ripple_solver_ArnoldiOrder2_test.f90
Fortran runtime error: Index '870' of dimension 1 of array 'npl' above upper bound of 869
#3  source_flux            at ripple_solver_ArnoldiOrder2_test.f90:4135
#4  ripple_solver_arnoldio2 at ripple_solver_ArnoldiOrder2_test.f90:2513
```

Observed (`Release`): the run writes `qflux_ntv.dat`, then aborts:

```
corrupted size vs. prev_size
Program received signal SIGABRT
#11 ripple_solver_arnoldio2 at ripple_solver_ArnoldiOrder2_test.f90:3371   (deallocate)
```

## Root cause

`npl` is allocated `(ibeg:iend)` (`ripple_solver_ArnoldiOrder2_test.f90:915` on `main`). Inside the internal `subroutine source_flux`, the second-neighbour coupling reads `npl` two steps out under only a one-step guard:

- `main:4061` `npassing_next = npl(istep + 2)`, inside `if (istep .LT. iend)` (`main:4005`). With `istep = iend-1` in the even branch, this reads `npl(iend+1)`.
- `main:3880` `npassing_next = npl(istep - 2)`, inside `if (istep .GT. ibeg)` (`main:3823`). With `istep = ibeg+1` in the even branch, this reads `npl(ibeg-1)`.

Whether the boundary `istep` lands in the even (`MOD(istep-ibeg,2)==0`) branch depends on the parity of `(iend-ibeg)`. For geometries where the field-line re-split makes that parity odd, `npl(iend+1)` / `npl(ibeg-1)` is accessed. The returned `npassing_next` is then used to index `q_rip(npassing_next+1:1:-1, istep+2, ...)` and to size `source_vector(...)` slices, so a garbage value produces the out-of-bounds **write** that corrupts the heap (only detected later, at the `deallocate` on line 3371).

## Pre-existing

Identical on `main`: `git show origin/main:NEO-2-QL/ripple_solver_ArnoldiOrder2_test.f90` has `npassing_next = npl(istep + 2)` at line 4061 and `npl(istep - 2)` at 3880. Not introduced by the recent QL PRs (#120/#121/#122/#119); those do not touch `source_flux`. It stays latent for geometries whose parity keeps the boundary `istep` in the odd branch (e.g. the AUG QL golden-record surfaces), and surfaces on helical-axis geometries after the `phi_divide>1` re-split.

## Proposed fix direction (owner)

Guard the second-neighbour reads so the index stays in `[ibeg, iend]` and supply the intended boundary value for `npassing_next` (mirroring how the first-neighbour terms are handled at the boundary), rather than relying on `(iend-ibeg)` parity. Alternatively, constrain the `phi_split_mode=3` re-split to preserve the parity the assembly assumes. This is `ripple_solver` core assembly; the boundary value feeds a physics coupling term, so it needs the solver maintainer's judgement, and any change must be checked against the QL golden records.

## Impact

Blocks QL NA runs on helical-axis (HELCORE l=1) geometries (the flux-pumping helical-core case). The zero-source guard in #122 handles the *sourceless* configuration; this is the distinct failure of a *driven* run on that geometry.

## Environment

- Branch `feat/ql-misalignment-drive` (line 4135); same code on `main` (line 4061).
- gfortran, Ninja/CMake build, single MPI rank.

krystophny:
Closes #123. This child is stacked on #125, #119, #121, and #124 and remains open against `main`.

## Summary

- keep both mode-3 field-line endpoints on full Runge-Kutta points
- when the requested subdivision count is odd, double the complete subdivision vector before constructing the refined grid
- test endpoint parity, original-node retention, ordering, and field-line reversal equivariance through the production divider

## Invariants

The repair leaves the second-order source stencils and their `5/12`, `2/3`, and `-1/12` weights unchanged. It preserves all requested points, local refinement ratios, coordinate orientation, trapped/passing mapping inputs, source signs, Fourier convention, CGS units, field-line boundary conditions, ABI, and storage layout. Even-count grids are bitwise unchanged. An odd-count retry grid uses twice the requested intervals, so both endpoints are full points and every `istep +/- 2` access remains inside the solve range.

The added resolution is limited to mode-3 retries whose requested interval count is odd. No branch in this stack is authorized to merge before the programme-wide gate.

### Test fails before fix

With the production divider unchanged:

```text
$ ctest --test-dir build -R '^phi_divider_parity_test$' --output-on-failure
0% tests passed, 1 tests failed out of 1
FAIL: refinement made the final Runge-Kutta point a half step
```

The production mapping for `[2,1,2,2]` ended at index 7. Original-node retention, ordering, and reversal checks already passed, isolating the endpoint contract.

### Test passes after fix

```text
$ fo
Static: OK (331 modules)
Build: OK
Tests: OK
Lint: OK
All stages passed

$ make test
100% tests passed, 0 tests failed out of 10

$ ctest --test-dir build -R '^phi_divider_parity_test$' --output-on-failure

$ git diff --check
<no output>
```

The odd request maps to `[4,2,4,4]`, ending at full-step index 14. The reversed request produces the reflected mapping, and every original point remains ordered and represented.

The 16 separately managed external-data regressions remain registered under `regression;external-data`; this branch does not mark that unavailable-data gate passed. A bounds-checked HELCORE production run and golden-record comparison remain integration gates.

### DIFF
--- a/COMMON/collision_operator_mems.f90
+++ b/COMMON/collision_operator_mems.f90
@@ -502,6 +502,7 @@ module collop
       if (allocated(denmm_a)) deallocate(denmm_a)
       if (allocated(Inbi_lmmp_a)) deallocate(Inbi_lmmp_a)
       if (allocated(asource)) deallocate(asource)
+      if (allocated(asource_hel)) deallocate(asource_hel)
       if (allocated(weightlag)) deallocate(weightlag)
       if (allocated(x1mm)) deallocate(x1mm)
       if (allocated(x2mm)) deallocate(x2mm)
--- a/COMMON/collop_compute.f90
+++ b/COMMON/collop_compute.f90
@@ -1507,6 +1507,7 @@ contains
   end subroutine compute_Minv
 
   subroutine compute_sources(asource_s, weightlag_s, weightden_s, weightparflow_s, weightenerg_s)
+    use rkstep_mod, only : asource_hel
     real(kind=dp), dimension(:,:), intent(out) :: asource_s, weightlag_s
     real(kind=dp), dimension(:), intent(out)   :: weightden_s, weightparflow_s, weightenerg_s
     real(kind=dp) :: res_int
@@ -1543,6 +1544,31 @@ contains
     end if
 
     call chop(asource_s)
+
+    ! Moments of the single-helicity misalignment drive (weights x^-1, x^2);
+    ! always computed by quadrature, no precomputed-matrix-element path.
+    if (allocated(asource_hel)) deallocate(asource_hel)
+    allocate(asource_hel(0:lagmax, 2))
+    if (isw_relativistic .eq. 0) then
+       do k = 1, 2
+          do m = 0, lagmax
+             asource_hel(m, k) = integrate(am_hel, 0d0)
+          end do
+       end do
+    else
+       do k = 1, 2
+          do m = 0, lagmax
+             asource_hel(m, k) = norm_maxwell * integrate(am_hel_rel1, 0d0)
+          end do
+       end do
+    end if
+    if (make_ortho) then
+       do k = 1, 2
+          asource_hel(:,k) = matmul(M_transform_inv, asource_hel(:,k))
+       end do
+    end if
+    call chop(asource_hel)
+
     write (*,*) "Computing weighting coefficients..."
 
     if (allocated(C_m)) deallocate(C_m)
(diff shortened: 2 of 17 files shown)


krystophny:

PR #132, opened 2026-07-11
Handle collision-kernel zero endpoints

krystophny:
## Summary

- use the analytic identity for the `l = 0` recurrence factor
- use the exact zero limit for the `l > 0`, `x = 0` inner kernel
- avoid invoking CQUAD for a callback that is identically zero on that endpoint interval
- add a halting-mode regression for the collision-operator zero endpoint

Stack dependency: #130. This PR remains pointed at `main` and must not merge before its dependency.

The change preserves the collision operator at every nonendpoint value. It does not change the quadrature algorithm, tolerances, discretization, convergence criteria, memory layout, ABI, or compiler semantics. At the endpoint, it evaluates the analytic limits: the `l = 0` recurrence multiplier is one, and the regular `l > 0` kernel is zero.

Closes #131.

## Verification

Failing before (`377226441848b64f39e97462937797e10342bab5`):

```text
0% tests passed, 1 tests failed out of 1
IEEE_INVALID_FLAG
FAIL: zero endpoint raised IEEE invalid
```

Passing after (`69fbb86`):

```text
100% tests passed, 0 tests failed out of 1
```

```text
Static analysis passed (335 Fortran files checked)
Build passed
100% tests passed, 0 tests failed out of 13
Lint passed
```

The focused regression completed without an IEEE invalid flag. `git diff --check` also passed.

PR #133, opened 2026-07-11
Skip exact-zero energy scattering integral

krystophny:
## Summary

- record whether the selected expansion basis has a constant zeroth mode
- return the exact zero energy-scattering matrix element for that mode at equal temperatures
- retain CQUAD for unequal temperatures and nonconstant modes
- extend the IEEE-halting regression with equal- and unequal-temperature controls

Stack dependency: #132 at `69fbb86dbce74526093b95936b7c052e4a7fccaf`. This PR remains pointed at `main` and must not merge before its dependency.

For a constant expansion mode, `d_phi_exp(0,x)=0`. At `T_a=T_b`, the remaining term is also identically zero, so the matrix element is exactly zero. The patch skips only that analytic case. It preserves every nonzero integrand, basis definition, quadrature algorithm and tolerance, discretization, convergence criterion, memory layout, ABI, and collision-operator units.

## Verification

Failing before (`69fbb86`):

```text
collop_zero_endpoint_test ... NUMERICAL
SIGFPE
COMMON/collop_compute.f90:1926
COMMON/gsl_integration_routines_mod.f90:647
```

GDB showed finite zero CQUAD samples with `ncdiff=0`, `nc=0` at `m=mp=0`, `T_a=T_b=1`; GSL evaluated `0/0` in its error heuristic.

Passing after (`03f5b9d`):

```text
collop_zero_endpoint_test ... Passed
```

```text
Static analysis passed (335 Fortran files checked)
Build passed
Tests passed
Lint passed
```

The regression also verifies that the unequal-temperature constant-mode element remains nonzero and is not skipped. A trap-enabled HELCORE run passes the former energy-scattering fault and reaches propagator joining; it then exposes a distinct zero-column normalization fault in `join_ripples.f90`, which is outside this patch.

### DIFF
--- a/COMMON/collision_operator_mems.f90
+++ b/COMMON/collision_operator_mems.f90
@@ -502,6 +502,7 @@ module collop
       if (allocated(denmm_a)) deallocate(denmm_a)
       if (allocated(Inbi_lmmp_a)) deallocate(Inbi_lmmp_a)
       if (allocated(asource)) deallocate(asource)
+      if (allocated(asource_hel)) deallocate(asource_hel)
       if (allocated(weightlag)) deallocate(weightlag)
       if (allocated(x1mm)) deallocate(x1mm)
       if (allocated(x2mm)) deallocate(x2mm)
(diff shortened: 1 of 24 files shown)


krystophny:

PR #135, opened 2026-07-11
Reject zero Lorentz join columns

krystophny:
## Summary

- validate Lorentz flux-conservation normalization factors before division
- report direction, column, factor, propagator tags, and pass dimensions
- return join error 5 for an exact-zero or nonfinite normalization factor
- extend the join diagnostic fixture with rejected-zero and accepted-finite cases

Stack dependency: #133 at `03f5b9dacdbf7c882fac01c37972e8e266575351`. This PR remains pointed at `main` and must not merge before its dependency.

For every nonzero finite column, the existing normalization and column-sum-one conservation law are unchanged. The patch does not choose a physical replacement for a zero column and does not change join algebra, linear solves, discretization, tolerances, array layout, ABI, or generated nonzero values. It converts an undefined `0/0` into an exact diagnostic rejection. The upstream meaning of the zero column remains tracked in #134.

## Verification

Failing before (`03f5b9d`), exact HELCORE source-only replay with IEEE-invalid trapping:

```text
join_ripples.f90:294
facnorm=0
SIGFPE: 0/0
```

Passing diagnostic after (`6e30459`):

```text
join_failure_diagnostic_test ... Passed
100% tests passed, 0 tests failed out of 1
```

```text
Static analysis passed (335 Fortran files checked)
Build passed
Tests passed
Lint passed
```

The same trapped replay now reports:

```text
join_ripples: normalization direction=p
join_ripples: normalization column=2
join_ripples: normalization factor=0.000000000000000E+00
join_ripples: old propagator tags= 12 13
join_ripples: new propagator tags= 13 13
join_ripples: pass dimensions= 24 24
Error from Joining, ierr=5
```

It exits without SIGFPE, IEEE invalid, NaN, `zhseqr`, or DGBSV failure. This is a diagnostic safety result, not a valid transport calculation; partial qflux files are rejected.

PR #136, opened 2026-07-11
Restrict Arnoldi solver to global route

### LINKED ISSUE #134: QL: Lorentz join normalizes an exact-zero propagator column
## Problem

The Lorentz flux-conservation correction in `NEO-2-QL/join_ripples.f90` divides every joined propagator column by its column sum without handling an exact-zero column.

At current stacked head `03f5b9dacdbf7c882fac01c37972e8e266575351`, a Debug build with bounds checking and `-ffpe-trap=invalid` reaches:

```text
join_ripples.f90:294
facnorm = 0
amat_p_p(1,2) = 0
amat_p_m(1,2) = 0
SIGFPE: 0/0
```

Context:

```text
i = 2
npass_l = 24
npass_r = 24
old propagator tags = 12..13
new propagator tags = 13..13
```

The collision operator and propagators complete before this fault. No earlier NaN, IEEE exception, or failed linear-solve diagnostic appears in the trapped run.

## Expected behavior

The flux-conservation correction must define and verify the intended behavior for a zero incoming propagator column. It must not silently create NaNs or divide by zero. Nonzero columns must retain the existing normalization and conservation law.

## Acceptance

- Add a behavioral fixture that reaches a joined exact-zero Lorentz column and fails before the repair.
- Define the physical meaning of the zero column: impossible state, removable empty channel, or upstream assembly defect.
- Preserve column-sum-one normalization for every nonzero column.
- Reject an impossible zero column with stage, index, dimensions, tags, and finite-value diagnostics, or handle a physically empty channel with an explicitly justified invariant.
- Run the full native suite and a bounds-checked source-only helical-geometry case.
- Verify no IEEE invalid flag and finite final transport outputs.

krystophny:
## Summary

- factor the existing one-propagator routing predicate
- reject ArnoldiO2 mode 3 before solve when geometry requires local propagator composition
- cover axisymmetric tokamak, homogeneous-field, and general-geometry routing cases

Stack dependency: #135 at `6e30459047a026f52da453eb9af59889107647a6`. This PR remains pointed at `main` and must not merge before its dependency.

ArnoldiO2 is a periodic global qflux solver. Existing mode-3 golden records all use the one-propagator route; the solver directly imposes endpoint phase conditions and never defines local transfer/source/flux outputs. Historical transfer construction was unreachable from the initial import and was deleted as dead code in 2019. Sending mode 3 through the all-propagator route therefore has no defined numerical meaning.

The patch preserves the existing routing predicate and every supported mode-3 calculation. It does not change the kinetic operator, periodic boundary conditions, sparse solver, tolerances, discretization, convergence, output layout, ABI, or any supported result. Mode 1 remains the local propagator solver for general geometry.

Closes #134.

## Verification

Failing before (`6e30459`), mode 3 with general helical geometry:

```text
join_ripples: normalization direction=p
join_ripples: normalization column=2
join_ripples: normalization factor=0
Error from Joining, ierr=5
```

The earlier unguarded stack performed `0/0` at the same undefined local column.

Passing after (`37d7076`):

```text
ripple_solver_route_test ... Passed
```

```text
Static analysis passed (337 Fortran files checked)
Build passed
Tests passed
Lint passed
```

The same trapped Debug run now exits before ArnoldiO2 or `join_ripples`:

```text
ERROR STOP isw_ripple_solver=3 requires the single-propagator global route
```

It produces no qflux output and reports no IEEE exception, NaN, SIGFPE, DGBSV failure, or partial transport result. This is configuration rejection, not a completed physics run.

### DIFF
--- a/COMMON/collision_operator_mems.f90
+++ b/COMMON/collision_operator_mems.f90
@@ -502,6 +502,7 @@ module collop
       if (allocated(denmm_a)) deallocate(denmm_a)
       if (allocated(Inbi_lmmp_a)) deallocate(Inbi_lmmp_a)
       if (allocated(asource)) deallocate(asource)
+      if (allocated(asource_hel)) deallocate(asource_hel)
       if (allocated(weightlag)) deallocate(weightlag)
       if (allocated(x1mm)) deallocate(x1mm)
       if (allocated(x2mm)) deallocate(x2mm)
--- a/COMMON/collop_compute.f90
+++ b/COMMON/collop_compute.f90
@@ -1507,6 +1507,7 @@ contains
   end subroutine compute_Minv
 
   subroutine compute_sources(asource_s, weightlag_s, weightden_s, weightparflow_s, weightenerg_s)
+    use rkstep_mod, only : asource_hel
     real(kind=dp), dimension(:,:), intent(out) :: asource_s, weightlag_s
     real(kind=dp), dimension(:), intent(out)   :: weightden_s, weightparflow_s, weightenerg_s
     real(kind=dp) :: res_int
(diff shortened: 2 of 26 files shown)


krystophny:

PR #140, opened 2026-07-11
Run helical source through local solver

krystophny:
Closes #139.

Parent stack head: `fix/flux-pumping/single-build-test` at `12aebf2` (PR #138). This cumulative PR targets `main` and remains unmerged with the rest of the feature stack.

Extract the existing complex Runge-Kutta helical-source stencil into one shared module. The periodic Arnoldi solver and mode-1 local solver call the same implementation. Mode 1 constructs the canonical complex band profiles from explicit source-only `m_phi_input`, then adds their real physical-space projection to its real right-hand side. The general-geometry path now returns its existing source and transfer matrices for propagator composition. The legacy per-propagator qflux diagnostic now iterates over the allocated species bounds instead of assuming two species.

Physics and numerical invariants:

- Fourier phase remains `exp(+i(m theta + n phi))`; the HELCORE memo `(1,1)` map is NEO-2 `(1,-1)`.
- CGS species quantities, Boozer coordinates, local boundary conditions, collision moments, Runge-Kutta stencil, passing-grid layout, kinetic operator, discretization, convergence criteria, transfer algebra, ABI, and generated-code semantics are unchanged.
- Corrugation remains sigma-odd; the electrostatic drive remains sigma-even. Only A1 and A2 source columns change.
- Mode 1 solves the real field. Real and imaginary potential amplitudes provide the two phase quadratures needed to reconstruct a complex harmonic response.
- Existing disabled-drive configurations remain covered by the full focused suite.

## Verification

Failing before:

```text
mode 3 general geometry: periodic Arnoldi solver requires the one-propagator global route
initial trapped mode-1 port: qflux_allspec third index 1 above upper bound 0
```

Passing after:

```text
$ fo
Static: OK (141 modules, 141 changed, 141 affected)
Build: OK
Tests: skipped, no affected tests
Lint: OK
All stages passed

$ fo test --verbose
Summary: 15 passed, 0 failed, 0 skipped
```

Focused source tests cover changing passing-grid sizes, forward and backward propagation, sigma parity, complex quadrature retention, force-column isolation, and pinned legacy stencil checksums.

Fresh Debug build flags:

```text
-fcheck=all -fbacktrace -g -O0 -ffpe-trap=invalid,zero,overflow
```

Exact source-only HELCORE `(m,n)=(1,-1)` result:

```text
exit 0; final join 24 x 24
wall 56.23 s; peak RSS 545304 KB
fulltransp gamma_out: finite 3 x 3
gamma current A1 = -7.33871
[...]

PR #142, opened 2026-07-11
Write dimensional helical response output

### LINKED ISSUE #141: Write a dimensional helical current-response contract
The local helical-source solver currently exposes only the legacy normalized `gamma_out` matrix. Add a machine-readable per-run response contract for the current channels without labeling a flux-surface moment as local current density.

Requirements:

- record the Fourier phase, mode numbers, surface, species, source amplitudes, solver route, and exact-zero/off-reference policy;
- expose `gamma_out(3,1:2)` as A1/A2 current-response channels;
- convert them with `D_3k = gamma_3k v_T rho B_ref`;
- use the normative units `cm^2 G/s` for D13/D23/D31/D32 and `cm^2 G^2/s` for D33;
- preserve the local kinetic operator, discretization, collision moments, convergence, transfer algebra, ABI, and memory layout;
- test the real/imaginary convention `H = (R-R0) - i(I-R0)` and require an off reference unless the zero-source guard proves the baseline exactly zero.

This contract represents the harmonic response coefficient. Local parallel current density, Boozer contravariant toroidal current, and geometric toroidal current require later coordinate/distribution reconstruction and must not be inferred from the scalar coefficient.

krystophny:
Closes #141.

Parent stack head: `feat/flux-pumping/local-helical-drive` at `47343f3` (PR #140). This cumulative PR targets `main` and remains unmerged with the rest of the feature stack.

Write `helical_current_response.h5` for single-species mode-1 helical-source runs. The contract records the input quadrature, Fourier phase, mode, surface, species, solver switches, source amplitudes, normalized current channels, dimensional D31/D32 coefficients, and whether a separate off reference is required. It does not label a flux-surface moment as local or geometric toroidal current.

Correct the existing multispecies HDF5 unit metadata to match the normative NEO-2 output document: D13/D23/D31/D32 use `cm^2 G/s`; D33 uses `cm^2 G^2/s`.

Numerical and physics invariants:

- `D_3k = gamma_3k v_T rho B_ref = gamma_3k 2 T c/(z e)`.
- Solver order and final-output mapping remain current row 3, A1/A2 columns 1/2.
- Fourier phase remains `exp(+i(m_theta theta + m_phi phi))`.
- The real/imaginary reconstruction is `H=(R-R0)-i(I-R0)`; source-only fileless runs use the existing exact-zero guard as R0=0.
- The kinetic operator, collision moments, discretization, convergence criteria, transfer algebra, ABI, and memory layout are unchanged.
- The output is a harmonic response coefficient. Pointwise parallel current and geometric toroidal current require a later distribution/coordinate reconstruction.

## Verification

Failing before:

```text
helical_current_response.h5: absent
D31/D32 HDF5 unit attribute: cm^2/s
```

Passing after:

```text
$ fo
Static: OK (340 modules, 340 changed, 340 affected)
Build: OK
Tests: OK
Lint: OK
All stages passed
```

The focused normalization/reconstruction, shared-source, and output-geometry tests pass 3/3. A three-case HELCORE integration completed every 24 x 24 final join in 3:24.90 with peak RSS 541996 KiB and no IEEE, NaN, signal, or runtime marker. The runner validated every HDF5 field, unit, phase, switch, species, finite value, and the agreement with `fulltransp` at relative tolerance `1e-13`.

Reconstructed `(m_theta,m_phi)=(1,-1)` response:

```text
D31 = 1.4677305578740989e12 - 5.42208046219255e11 i cm^2 G/s
D32 = 4.18887130153984e12 - 6.277850811343069e11 i cm^2 G/s
```

The exact-zero source guard supplies the baseline for this fileless source-only case.

### DIFF
--- a/COMMON/collision_operator_mems.f90
+++ b/COMMON/collision_operator_mems.f90
@@ -502,6 +502,7 @@ module collop
       if (allocated(denmm_a)) deallocate(denmm_a)
       if (allocated(Inbi_lmmp_a)) deallocate(Inbi_lmmp_a)
       if (allocated(asource)) deallocate(asource)
+      if (allocated(asource_hel)) deallocate(asource_hel)
       if (allocated(weightlag)) deallocate(weightlag)
       if (allocated(x1mm)) deallocate(x1mm)
       if (allocated(x2mm)) deallocate(x2mm)
--- a/COMMON/collop_compute.f90
+++ b/COMMON/collop_compute.f90
@@ -1507,6 +1507,7 @@ contains
   end subroutine compute_Minv
 
   subroutine compute_sources(asource_s, weightlag_s, weightden_s, weightparflow_s, weightenerg_s)
+    use rkstep_mod, only : asource_hel
     real(kind=dp), dimension(:,:), intent(out) :: asource_s, weightlag_s
     real(kind=dp), dimension(:), intent(out)   :: weightden_s, weightparflow_s, weightenerg_s
     real(kind=dp) :: res_int
@@ -1543,6 +1544,31 @@ contains
     end if
 
     call chop(asource_s)
+
+    ! Moments of the single-helicity misalignment drive (weights x^-1, x^2);
+    ! always computed by quadrature, no precomputed-matrix-element path.
+    if (allocated(asource_hel)) deallocate(asource_hel)
+    allocate(asource_hel(0:lagmax, 2))
+    if (isw_relativistic .eq. 0) then
+       do k = 1, 2
+          do m = 0, lagmax
+             asource_hel(m, k) = integrate(am_hel, 0d0)
+          end do
+       end do
+    else
+       do k = 1, 2
+          do m = 0, lagmax
+             asource_hel(m, k) = norm_maxwell * integrate(am_hel_rel1, 0d0)
+          end do
+       end do
+    end if
+    if (make_ortho) then
+       do k = 1, 2
+          asource_hel(:,k) = matmul(M_transform_inv, asource_hel(:,k))
+       end do
+    end if
+    call chop(asource_hel)
+
     write (*,*) "Computing weighting coefficients..."
 
     if (allocated(C_m)) deallocate(C_m)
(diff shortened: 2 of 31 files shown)


krystophny:
### Added slice: pointwise helical parallel-current profile (commit 6439ac9)

Extends the resolved-current output with a per-field-line-point current profile
for the mode-1 local (axi) solver.

**What it does.** After the transport contraction
`qflux(2,k) = sum_col flux_vector(2,col)*source_vector(col,k)`, the new
`pointwise_current_mod` decomposes that current channel into its per-point
contributions. `flux_vector`'s columns partition into disjoint per-step blocks,
so the block dot products sum back to `qflux(2,k)` exactly. The routine reuses
the already-verified current row: the counter-passing block carries the explicit
minus sign (parallel current is odd under sigma -> -sigma) and the phi-quadrature
weight is baked in, so no new sign, weight, or normalization is introduced.

**Invariant preserved.** The solver writes `current_profile.dat` only when the
helical drive is active (`isw_hel_drive /= 0`), so runs without it are bitwise
unchanged. A hard self-check aborts the run if the field-line sum does not
reproduce `qflux(2,k)`.

**Verification.**
- `ctest` TEST suite: 17/17 pass; new `pointwise_current_test` added.
- Red/green: breaking the block partition (dropping the `npassing+1` factor)
  aborts with `ERROR STOP FAIL: per-point contributions do not sum to the qflux
  current channel`; restored -> Passed.
- Single-surface l=1 helical-core benchmark, three drive quadratures: the
  harmonic current response reproduces the reference to ~1e-11, and each
  `current_profile.dat` reconstructs `qflux(2,k)` to ~1e-15 relative over 961
  finite points.

Parent slice: dimensional helical response contract (this PR's earlier commits).
No merge requested; review only under the integration freeze.

krystophny:
Correction: the pointwise current-profile slice was moved to its own stacked PR #151 (top of the stack, on the source-conjugacy branch) to keep this PR focused on the dimensional response contract and to avoid rebasing the intervening open PRs. Please disregard the previous comment's commit reference; the pointwise work is reviewed in #151.

krystophny:

PR #153, opened 2026-07-11
Reject numerical ties in field-line return selection

krystophny:
## Summary

- require a return distance to improve by more than the floating-point near-tie threshold
- preserve the first materially closer periodic return
- cover the observed near-tie and true return with a unit test

The change leaves field integration, return-distance evaluation, propagator construction, ABI, and memory layout unchanged. It only changes the comparison used to select the terminating return.

## Verification

Failing before, with fixed geometry and increasing steps per period:

```text
nstep=1920: period=100, distance=3.8516034805979871E-10
nstep=3840: period=99, distance=6.2831849552580366E-02
period-1 minimum: 6.2831853072124133E-02
```

The 3840-step result selected a numerical near-tie that was 3.5E-9 smaller than the period-1 distance.

Passing after:

```text
return_distance_test PASS 0.10s

Static: OK
Build: OK
Tests: OK
Lint: OK
All stages passed
```

The formatter reports the existing legacy whole-file style warning for `COMMON/mag_interface.f90`; no broad reformat is included.


krystophny:

PR #154, opened 2026-07-13
Record propagator join normalization factors

krystophny:
## Summary

- optionally record each finite Lorentz-model `join_ripples` normalization factor before division
- include sequence, direction, column, old and new propagator tags, and passing dimensions in CSV output
- leave output disabled unless `NEO2_JOIN_NORMALIZATION_FILE` is set
- fail explicitly if requested diagnostic output cannot be created or written

The diagnostic does not change propagator matrices, normalization factors, flux-conservation division, solver dispatch, convergence criteria, ABI, or memory layout.

## Verification

Failing before:

```text
$ ctest --test-dir build -R '^join_failure_diagnostic_test$' --output-on-failure
ERROR STOP FAIL: join normalization output is absent
0% tests passed, 1 tests failed out of 1
```

Passing after:

```text
$ ctest --test-dir build -R '^join_failure_diagnostic_test$' --output-on-failure
100% tests passed, 0 tests failed out of 1

$ fo
Static: OK (347 modules, 347 changed, 347 affected)
Build: OK
Tests: OK
Lint: OK
All stages passed (.9s)
```

PR #157, opened 2026-07-13
Record active phi-grid placement decisions

krystophny:
## Summary

This PR adds an opt-in record of the phi-grid strategy that `modify_propagator` actually applies.

The first implementation recorded only automatic `phi_placer` intervals. A production reconstruction then emitted no CSV because its configured path uses `phi_split_mode=1`; the automatic eta placer is never called. The recorder now covers the decision boundary shared by all modes:

- `placed`: automatic eta-interval placement, with the raw `CEILING` decision and the count after the existing odd/minimum adjustment;
- `halfstep`: configured mode-1 refinement, with input and output grid-point counts;
- `no_eta_halfstep`: the mode-2 fallback when no eta value lies in the propagator interval;
- `divider`: the mode-3 retry path, with input and output grid-point counts.

Each row identifies the propagator and outcome. It also records the interval, branch, eta index, phi endpoints, and original phi spacing. The generic count columns are named `decision_count` and `result_count` because their exact meaning is strategy-specific.

## Behavior and invariants

Set `NEO2_PHI_PLACEMENT_FILE` to request the CSV. With the variable unset, the recorder performs no allocation or file I/O. A requested path that cannot be written is a hard error rather than an incomplete record.

The diagnostic does not change the selected strategy, `CEILING` expression, parity adjustment, half-step or divider grid, point locations, field-line integration, source, collision operator, join normalization, convergence criteria, ABI, or generated physics-data layout. It only observes the grid before `modify_propagator` replaces the old coordinates.

## Relation to the flux-pumping stack

This branch is stacked on PR #154 because the downstream convergence campaign uses one pinned executable history. It does not depend on the join-normalization values. The record is intended to distinguish a response change caused by a phi-grid strategy or count transition from a downstream change.

## Verification

Failing before the recorder existed:

```text
phi_placement_diagnostic_test: record_phi_placement is absent
```

The first production attempt at the previous head also failed strict validation because no requested CSV was created. That exposed the process-coverage gap: the unit test exercised the writer, but the carrier never entered automatic placement.

Focused test at `defe172`:

```text
phi_placement_diagnostic_test  PASS  0.06s
```

Full repository pipeline:

```text
Static: OK (348 modules, 348 changed, 348 affected)
Build: OK
Tests: OK
Lint: OK
All stages passed
Fmt: WARN
```

The format warning is existing whole-file drift in `COMMON/flint_mod.f90` and `NEO-2-QL/flint.f90`; the touched lines and staged diff pass whitespace and line-length checks.

A real three-stage reconstruction at `nstep=480` now writes 203 validated rows in stage 0 and 203 in stage 2. The records are byte-identical between those stages: 199 use `halfstep` and four use the `divider` retry path. Decision counts span 5-865 grid points and result counts span 9-1739. The solve retains 199 reconstructed responses and closes qflux below `8.7e-15` relative error.

### DIFF
--- a/COMMON/flint_mod.f90
+++ b/COMMON/flint_mod.f90
@@ -26,11 +26,12 @@ MODULE flint_mod
 
 contains
 
-  SUBROUTINE record_phi_placement(tag,interval,part,eta_index,phi_start, &
-       phi_end,hphi,ceiling_count,placed_count,ierr)
+  SUBROUTINE record_phi_placement(tag,outcome,interval,part,eta_index, &
+       phi_start,phi_end,hphi,decision_count,result_count,ierr)
     INTEGER, INTENT(in) :: tag,interval,part,eta_index
-    INTEGER, INTENT(in) :: ceiling_count,placed_count
+    INTEGER, INTENT(in) :: decision_count,result_count
     DOUBLE PRECISION, INTENT(in) :: phi_start,phi_end,hphi
+    CHARACTER(len=*), INTENT(in) :: outcome
     INTEGER, INTENT(out) :: ierr
 
     INTEGER :: iunit,status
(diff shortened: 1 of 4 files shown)


krystophny:

PR #158, opened 2026-07-13
Fix collision precompute initialization and CQUAD IEEE handling

krystophny:
## Summary

This PR fixes two initialization failures exposed by a trap-enabled multispecies QL run with B-spline collision bases.

1. Collision-table precomputation runs before a physical species pair is selected. The precomputed inner kernels are mass-independent, but finite-interval subdivision previously evaluated `log(m_b/m_a)` while both module values were undefined. The PR initializes that precompute state to the equal-mass subdivision value and uses one subdivision helper in both finite-interval integrators.
2. After that fix, GSL CQUAD raised IEEE invalid on exact-zero B-spline subintervals even though every callback value and the returned integral were finite. Both parameterless and one-parameter CQUAD wrappers now isolate that library-internal flag. Each callback clears and checks IEEE invalid around the user integrand, and records nonfinite callback values. The wrapper restores the caller's prior IEEE flag and halting mode, then rejects any integrand exception or nonfinite CQUAD result.

Divide-by-zero and overflow trapping remain enabled throughout. Invalid arithmetic in a collision kernel or any other user integrand remains fatal.

## Physics and numerical invariants

The change preserves:

- collision kernels, B-spline bases, knot locations, integration intervals, tolerances, and CQUAD results;
- multispecies mass assignment before physical collision-operator construction;
- particle and energy conservation corrections, source construction, and matrix dimensions;
- convergence criteria, MPI/OpenMP behavior, ABI, and memory layout.

The intentional semantic change is limited to IEEE state: a finite successful CQUAD call no longer leaks an invalid flag generated inside CQUAD. Invalid or nonfinite physics callbacks still terminate the run.

## Relation to the flux-pumping stack

This branch is stacked on PR #154 because the downstream diagnostic campaign uses that pinned executable history. The collision fix is independent of the join-normalization recorder and does not change the helical drive, source assembly, response normalization, or aligned-potential definition. It unblocks the rotating aligned-potential regression used to test cancellation of the corrugation-only current response.

## Verification

Failing before the mass initialization:

```text
collop_precompute_mass_test  ***Exception: Numerical
collop_compute.f90:1036
Program received signal SIGFPE
```

Failing after the mass initialization but before the CQUAD guard:

```text
collop_bspline_precompute_test  FAIL  0.00s
Program received signal SIGFPE
gsl_integration_routines_mod.f90:724
collop_compute.f90:2648
```

Focused tests after both fixes:

```text
collop_precompute_mass_test       PASS  0.07s
collop_zero_endpoint_test         PASS  0.12s
cquad_ieee_guard_test             PASS  0.14s
collop_bspline_precompute_test    PASS  1.37s
Summary: 4 passed, 0 failed, 0 skipped
```

`cquad_ieee_guard_test` checks both sides of the boundary: an exact-zero integrand returns zero without leaking IEEE invalid, while a child process with an invalid integrand must fail.

The trap-enabled production carrier completes without an IEEE or runtime marker. Its six reported transport channels agree with the previous non-halting executable to a maximum relative difference of `9.82e-14`.

The full rotating aligned-potential regression also passes:

```text
aligned suppression = 393.1442870775852
sign-reversed ratio = 2.00254359538236
acceptance = true
```

Full repository pipeline:

```text
Static: OK (350 modules, 350 changed, 350 affected)
Build: OK
Tests: OK
Lint: OK
Fmt: WARN
All stages passed
```

The format warning is existing whole-file drift in `gsl_integration_routines_mod.f90`; the touched lines and complete staged diff pass whitespace and line-length checks. No broad legacy reformat is included.

### DIFF
--- a/COMMON/collop_compute.f90
+++ b/COMMON/collop_compute.f90
@@ -56,8 +56,8 @@ module collop_compute
   !**********************************************************
   ! Species
   !**********************************************************
-  real(kind=dp)    :: m_a
-  real(kind=dp)    :: m_b
+  real(kind=dp)    :: m_a = 1.0d0
+  real(kind=dp)    :: m_b = 1.0d0
   character(len=3) :: tag_a
   character(len=3) :: tag_b
 
@@ -955,7 +955,7 @@ contains
 
        if (lsw_split_interval) then
           do k = 1, kmax-1
-             n_sub = num_sub_intervals*max(int(log(m_b/m_a)), 1)
+             n_sub = scaled_num_sub_intervals()
              if ((.not. binknots) .and. (k .eq. kmax-1)) n_sub = num_sub_intervals_cutoff
              x_sub_del = (x_inter(k+1) - x_inter(k))/dble(n_sub)
              do k_sub = 1,n_sub
@@ -1033,7 +1033,7 @@ contains
 
        if (lsw_split_interval) then
           do k = 1, kmax-1
-             n_sub = num_sub_intervals*max(int(log(m_b/m_a)), 1)
+             n_sub = scaled_num_sub_intervals()
              if ((.not. binknots) .and. (k .eq. kmax-1)) n_sub = num_sub_intervals_cutoff
              x_sub_del = (x_inter(k+1) - x_inter(k))/dble(n_sub)
              do k_sub = 1,n_sub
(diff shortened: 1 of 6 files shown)


krystophny:

PR #161, opened 2026-07-13
Trace periodic join compatibility and nullspace

krystophny:
`join_ends` replaces one periodic equation per retained Laguerre block with the Lorentz current-closure constraint. The remaining periodic seam could therefore have come from an incompatible source, an incorrect nullspace, or reconstruction of a different source.

Add an opt-in `NEO2_JOIN_END_TRACE_FILE` diagnostic. For each join it records the Lorentz source correction, transfer-matrix identity error, original periodic-system left and right null vectors, compatibility residual, both replaced-row residuals, and algebraic scales. The CSV writer rejects wrong shapes, nonfinite values, nonpositive scales, and I/O failures.

The diagnostic is observation-only. When disabled it allocates no diagnostic arrays and performs no diagnostic solve. It does not change the Lorentz projection, Pfirsch-Schlueter current constraint, particle-conservation treatment, collision operator, discretization, convergence criteria, Fourier convention, CGS units, boundary conditions, ABI, or array layout.

The accepted 480-step, 30-pitch batch replay resolves the open hypothesis:

- periodic compatibility residual: `2.81e-17` relative;
- replaced p/m row residuals: `1.18e-14` and `1.17e-14` relative;
- replaced-row difference: `1.39e-16` relative;
- left/right null residuals: `1.39e-16` and `5.27e-15` relative;
- Lorentz projection identity residual: `2.71e-16` relative;
- Lorentz source correction: `7.85e-4` of the unsigned source scale;
- periodic reconstructed-distribution seam: `7.28e-4`;
- periodic current-contraction seam: `8.48e-4`.

The periodic solve is compatible and satisfies both original rows. The seam instead follows from applying the Lorentz correction only to the joined propagator source. Stage-1 reconstruction replays the saved local sources, which do not contain a spatially localized form of that correction. This PR does not invent such a localization and does not accept a physical pointwise current.

With the diagnostic enabled, all 398 saved propagators and 199 reconstructed boundaries are bit-for-bit identical to the parent result. Every numerical dataset in the 199 local-response HDF5 files and the global response file is also identical. All 12 focused tests and the trace-analysis test pass under bounds checks and floating-point traps. A follow-up commit separates finiteness and positivity checks because Fortran logical operators are not required to short-circuit; this prevents a nonfinite diagnostic scale from reaching an ordered comparison under invalid-operation traps.

### Test fails on parent

At `130ea2c`:

```text
rm -f build/TEST/join_end_trace.csv
fo test join_failure_diagnostic_test
test -s build/TEST/join_end_trace.csv

join_failure_diagnostic_test PASS
focused_test_status=0
trace_file_status=1
```

The existing test passes, but no requested join-end trace is produced.

### Test passes after fix

```text
rm -f build/TEST/join_end_trace.csv
fo test join_failure_diagnostic_test
test -s build/TEST/join_end_trace.csv

join_failure_diagnostic_test PASS
focused_test_status=0
trace_file_status=0
```

```text
fo

Static: OK
Build: OK
Tests: OK
Lint: OK
Fmt: WARN
All stages passed
```

The format warning is already present for the legacy `join_ends.f90` file at the parent commit. This PR keeps its existing layout instead of reformatting unrelated lines.

### DIFF
--- a/NEO-2-QL/join_diagnostics_mod.f90
+++ b/NEO-2-QL/join_diagnostics_mod.f90
@@ -182,13 +182,16 @@ contains
             .or. .not. all(ieee_is_finite(right_residual)) &
             .or. .not. all(ieee_is_finite(left_scale)) &
             .or. .not. all(ieee_is_finite(right_scale)) &
-            .or. any(compatibility_scale <= 0.0_real64) &
+            .or. .not. all(ieee_is_finite(transfer_error))) then
+            ierr = 7
+            return
+        end if
+        if (any(compatibility_scale <= 0.0_real64) &
             .or. any(dropped_p_scale <= 0.0_real64) &
             .or. any(dropped_m_scale <= 0.0_real64) &
             .or. any(left_scale <= 0.0_real64) &
             .or. any(right_scale <= 0.0_real64) &
-            .or. any(source_scale <= 0.0_real64) &
-            .or. .not. all(ieee_is_finite(transfer_error))) then
+            .or. any(source_scale <= 0.0_real64)) then
             ierr = 7
             return
         end if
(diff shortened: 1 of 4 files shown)


krystophny:

PR #163, opened 2026-07-13
Trace local Lorentz constant state

krystophny:
## Summary

Add an opt-in stage trace that evaluates the exact pitch-independent Lorentz state at three points in every single-ripple propagator:

1. against the assembled sparse equations before factorization;
2. after one direct sparse solve with exact constant boundary data;
3. in the extracted boundary map already traced by #162.

This PR is stacked on #162. It is the minimum measurement needed to localize the finite-resolution constant-distribution transport defect before changing the solver.

The internal state uses the code's documented unknown

\[
Y_k=\int_{\eta_{k-1}}^{\eta_k}f\,d\eta,
\]

including the moving final-band width \(1/B-\eta_N\) and the reversed storage order of the counter-passing branch. The sparse right-hand side applies those exact widths only at the incoming co-passing and counter-passing boundaries. The matrix residual and solved-state error are appended to `NEO2_LOCAL_PROJECTION_TRACE_FILE`; with that variable unset, the new path returns before allocation and does not call the extra solve.

## Decision rule

- Nonzero sparse residual: discretization or mirror matching is inconsistent with constant \(f\).
- Sparse residual closes but solved state does not: sparse factorization or solve is responsible.
- Solved state closes but #162's boundary map does not: propagator extraction is responsible.

No correction is applied in this PR. The measured stage will determine the smallest justified fix.

## Preserved invariants

- Lorentz collision and physical source terms are unchanged.
- Pitch and spatial discretizations, solver tolerances, and convergence criteria are unchanged.
- Fourier convention, CGS units, periodic streaming, and mirror boundary conditions are unchanged.
- ABI, serialized propagator layout, and generated physics outputs are unchanged.

### Test fails on main

Main has no executable check for the exact internal constant state or its sparse residual:

```text
Test project .../build
No tests were found!!!
Errors while running CTest
```

### Test passes after fix

```text
$ fo test lorentz_projection_diagnostic_test
lorentz_projection_diagnostic_test PASS 0.06s

$ fo
Static: OK
Build: OK
Tests: OK
Lint: OK
Fmt: WARN (legacy ripple_solver_axi_test.f90)
All stages passed
```

## Diagnosis and successor

The stage trace localized the defect to a topology crossing, not the sparse
factorization, boundary-map extraction, pitch transfer, joining, or the Lorentz
collision stencil. In propagator 192, the number of passing bands falls from 27
to 26 between spatial steps 679 and 680. The old placement routine does not
align the crossing sample when that sample is closer than both neighbours. The
constant-distribution widths on the two sides then differ by
`3.365342566301611e-4`; multiplication by the streaming coefficient produces
the complete `5.1395606791432513e-2` sparse row residual. Collision terms cancel
on the same row.

PR #164 supplies the stacked correction and regression. At its pinned head,
the worst normalized sparse constant residual is `1.0147e-16`, the worst solved
constant-state error is `3.0437e-14`, and the full 199-propagator reconstruction
completes with the same eight adaptive retries as the stacked base. A retained
stage-0 trace closes right transport at `2.2289e-14` and projector intertwining
at `2.3275e-14`. The failure-free campaign remains open. This PR remains
diagnostic-only.

### DIFF
--- a/NEO-2-QL/lorentz_projection_diagnostics_mod.f90
+++ b/NEO-2-QL/lorentz_projection_diagnostics_mod.f90
@@ -316,24 +316,33 @@ contains
     end subroutine record_local_constant_stage_residuals
 
     subroutine record_local_constant_row(tag, row, irow, icol, values, state, &
-            rhs, ierr)
+            rhs, npl, ind_start, bhat, ibeg, iend, lag, ierr)
         integer, intent(in) :: tag, row, irow(:), icol(:)
         real(real64), intent(in) :: values(:), state(:), rhs(:)
+        integer, intent(in) :: npl(ibeg:), ind_start(ibeg:)
+        real(real64), intent(in) :: bhat(ibeg:)
+        integer, intent(in) :: ibeg, iend, lag
         integer, intent(out) :: ierr
         character(len=:), allocatable :: filename
-        integer :: entry, iunit, status
+        integer :: column_band, column_laguerre, column_sigma, column_step
+        integer :: entry, iunit, row_band, row_laguerre, row_sigma, row_step, status
 
         ierr = 1
         if (.not. allocated(output_filename)) return
         if (size(irow) /= size(icol) .or. size(irow) /= size(values)) return
         if (row < 1 .or. row > size(state) .or. size(rhs) /= size(state)) return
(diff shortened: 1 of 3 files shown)


krystophny:

PR #165, opened 2026-07-14
Preserve geometry at pitch crossings

krystophny:
## Problem

PR 164 makes the discrete constant-distribution null exact, but its crossing
repair changes only `bhat_mfl` at a grid point. The coordinate, metric, drift,
and derivative arrays remain sampled at the old point. This creates an
internally inconsistent geometry that moves with spatial resolution.

## Change

- Locate each simple trapped-passing boundary from the bracketed root
  `1 - eta B(phi) = 0` in the existing six-point Lagrange representation.
- Move the nearer interior node to the root without changing the node count,
  pitch bands, array bounds, ABI, or serialized propagator layout.
- Interpolate `B`, `h_phi`, `dlogB/ds`, `bcovar_s_hat`, and `geodcu` from one
  unchanged stencil. Recompute `dlogB/dphi` and `d(bcovar_s_hat)/dphi` from
  the same polynomial.
- Apply the same path to the Arnoldi and axisymmetric QL solvers. Derived NTV
  arrays and nonuniform cell factors are built after alignment.
- Request spatial refinement when the bracket is lost, the root is not simple,
  one cell crosses more than one pitch band, the field is not monotone away
  from its minimum, or two crossings require the same node.

The magnetic-field representation, CGS units, pitch measure, mirror and
periodic endpoint conventions, node order, and solver storage layout are
unchanged. This PR changes where existing field data are sampled; it does not
change the collision operator or introduce a new source.

This PR is stacked on #164 and should remain unmerged until the pinned spatial
and pitch campaigns pass.

### Test fails on the base branch

With the new behavioral test applied before the implementation:

```text
$ fo test phi_crossing_alignment_test
TEST/test_phi_crossing_alignment.f90: Function 'claim_crossing_target' has no IMPLICIT type
Summary: build failed
```

### Test passes after fix

```text
$ fo test phi_crossing_alignment_test
Summary: 1 passed, 0 failed, 0 skipped (  0.06s)
```

```text
$ fo
Static: OK (356 modules, 356 changed, 356 affected)
Build: OK
Tests: OK
Lint: OK
Fmt: WARN
All stages passed (.7s)
```

The format warning is limited to the two pre-existing legacy solver files;
the module and behavioral test are formatted.

### DIFF
--- a/NEO-2-QL/phi_crossing_alignment_mod.f90
+++ b/NEO-2-QL/phi_crossing_alignment_mod.f90
@@ -1,7 +1,20 @@
 module phi_crossing_alignment_mod
     use, intrinsic :: iso_fortran_env, only: dp => real64
+    use plagrange_mod, only: plagrange_coeff
 
     implicit none
+
+    private
+
+    integer, parameter, public :: crossing_ok = 0
+    integer, parameter, public :: crossing_no_bracket = 1
+    integer, parameter, public :: crossing_non_simple = 2
+    integer, parameter, public :: crossing_invalid_stencil = 3
+    integer, parameter :: crossing_stencil_size = 6
+
+    public :: bracketed_phi_crossing
+    public :: claim_crossing_target
+    public :: nearest_phi_crossing_offset
 
 contains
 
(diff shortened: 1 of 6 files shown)


krystophny:
Spatial campaign 18573023 completed at 1,920/3,840/7,680 steps on the pinned PR head. All three component runs exited zero with no adaptive recoveries; topology stayed fixed and the maximum qflux relative error was 3.48e-14. The two harmonic-response components were (-13.6163181,-70.1506689), (-13.6152840,-70.1478445), and (-13.6134972,-70.1396413). Successive relative changes were (7.59e-5,4.03e-5) and (1.31e-4,1.17e-4): both are below 1e-3, but the final change grows rather than contracts, so this does not pass the spatial convergence gate.

The submission also omitted the local-projection and join-end trace flags. It therefore did not record the constant-distribution closure or global Lorentz correction at each resolution and is retained only as a rejected diagnostic. Campaign 18573042 repeats 3,840/7,680 and extends to 15,360 steps with both traces enabled. The incomplete pitch campaign was cancelled and replaced by diagnostic-complete campaign 18573043 at 30/60/120/240 bands. Collision physics, source, coordinates, units, boundary conditions, ABI, and serialized layout remain unchanged. No pointwise current or physical profile is accepted pending both corrected gates.

krystophny:
## Bounds/gate defect fixed in 6aec6fc

### Defect

`rearrange_phideps` passed `ubound(phi_divide, 1)` to
`align_phi_crossing_geometry` as the physical scan bound. The dummy
`phi_divide` was declared `dimension(1:iend)` with
`iend = ub_mag + 2*modify_br`, while both callers allocate
`phi_divide(1:ub_mag)`. With right boundary-layer modification active
(`modify_br = 1`, reachable through `ripple_solver_ArnoldiO2`; the
axisymmetric solver pins `modify_br = 0` via its long-standing
`goto 10`), three things went wrong:

1. The crossing scan ran over `0:ub_mag+2` and ingested the two
   duplicated padding nodes, so `bracketed_phi_crossing`'s
   strict-monotonicity guard (`phi(iend-1) == phi(iend)`) rejected
   every crossing regardless of location - each propagator with a
   pitch crossing requested refinement forever.
2. A padding-region rejection clamped `failed_interval` with
   `min(ub_mag+2, ...)` and wrote the refinement flag past the actual
   `phi_divide(1:ub_mag)` extent.
3. The caller's `maxval(phi_divide) > 1` gate never saw that flag, so
   the `ierr = 3` retry was silently lost and the solver continued
   with uninitialized `delt_pos/delt_neg/fact_*`.

The pinned HELCORE campaigns (18573042/18573043) ran the axisymmetric
solver with `modify_br = 0` (`iend == ub_mag`), so their recorded
results are unaffected.

### Fix

- Pass the physical `ub_mag` into `rearrange_phideps` explicitly and
  forward it to `align_phi_crossing_geometry`; the crossing scan now
  covers `0:ub_mag` on both sides of the field minimum.
- Size the `phi_divide` dummy as `(1:ub_mag)`, matching every actual
  argument, so the blanket `phi_divide = 1` and the rejection flag stay
  in bounds and the caller gate fires for every rejection.
- Clamp the legacy sub-interval split writes to
  `max(1, ib+1):min(ie, ub_mag)`; this also closes the pre-existing
  below-bounds write for `modify_bl = 1` (`ib + 1 = ibeg + 1 = -1`).

With `modify_br = 0` the passed bound equals `iend`, all clamps are
identities, and the arithmetic is unchanged - the pinned campaign
behavior is bit-identical.

### Verification

New behavioral test `TEST/test_crossing_geometry_padding.f90`
(`crossing_geometry_padding_test`): a 41-node parabolic field with one
pitch band and two simple crossings, run unpadded and with
`modify_br = 1` padding (boundary value above all physical values, and
below the crossing band so the padding mimics an extra crossing), plus
a padded case with a non-monotone physical field. It requires bitwise
agreement of the padded and unpadded aligned geometry, no spurious
refinement, untouched padding nodes, consistent integration steps, and
the rejection flag inside `phi_divide(1:ub_mag)`.

Test fails on this PR's previous head 0345e33 (call adapted to the old
signature, implicit interface exactly as the solver callers use):

```text
$ fo test crossing_geometry_padding_test
crossing_geometry_padding_test             FAIL       0.00s
ERROR STOP FAIL: benign padding triggered spurious refinement
```

Test passes after the fix, along with the existing crossing tests:

```text
$ fo test crossing_geometry_padding_test phi_crossing_alignment_test phi_divider_parity_test phi_placement_diagnostic_test
phi_placement_diagnostic_test              PASS       0.06s
phi_divider_parity_test                    PASS       0.07s
crossing_geometry_padding_test             PASS       0.06s
Summary: 4 passed, 0 failed, 0 skipped (  0.25s)

$ fo
Build: OK
Tests: OK
Lint: OK
All stages passed (51.1s)
```

ctest: 43/44 pass; the single failure is the pre-existing
external-data case `Extended_QL_Multispecies_Read` (needs
`NEO2_TEST_PATH` data not present locally), excluded from the local
pipeline by the external-data classification.

NEO-2-PAR keeps its own separate `rearrange_phideps` (old signature,
same latent extent mismatch); porting the alignment there or
documenting the divergence stays on the consolidation list.

krystophny:

PR #167, opened 2026-07-16
Retain per-level diagnostics for closure attribution

krystophny:
## Problem

The rejected PR 165 campaigns localized the 8.05e-4/7.61e-4 interface jump to
the periodic-closure pair and attributed it to the Lorentz solvability
projection that `join_ends` applies only to the final joined propagator, while
the stage-2 local re-solves use unprojected sources. That attribution is
plausible but unproven (2.6-3.1% magnitude mismatch), and the recorded runs
left three observability gaps: the projection footprint was not reconstructible
per band, adaptive pitch recoveries were only countable rather than
attributable, and the 20.5 GiB campaign peak RSS had no per-process source.
Separately, `lorentz_projection_diagnostics_mod` reported trace I/O failures as
ierr=3, which `propagator.f90` treats as a retryable phi-refinement request.

## Change

- Export the per-band, per-force corrections `delta_eta_l/r * facnorm` that
  `join_ends` subtracts from `source_m`/`source_p`, as additive kinds
  `delta_eta_l`, `delta_eta_r`, `source_correction_m`, `source_correction_p`
  in the existing join-end trace CSV schema.
- Add the discriminator experiment for the closure jump: when
  `NEO2_STAGE2_ATTRIBUTION_FILE` names a footprint file (rows
  `tag,side,force,band,value` taken from the stage-0 export above), the
  stage-2 re-solve subtracts those band vectors from the **assembled**
  right-hand side before `sparse_solve`, mapped onto the local propagator
  phase-space grid (side p: `ind_start(iend)+band`; side m:
  `ind_start(ibeg)+2*(npl(ibeg)+1)+1-band`), not taken from the post-solve
  `source_vector_all`. Applied rows go to `<file>.applied`; the remaining
  jump stays a measured quantity downstream — the footprint identity is not
  assumed.
- Add `peak_rss_mod`: per-process VmHWM from `/proc/self/status`, written to
  `NEO2_PEAK_RSS_FILE` following the `NEO2_*_FILE` pattern of
  `qflux_profile_mod`, hooked at end of run in `neo2.f90`.
- Report projection-trace I/O failures as the distinct public code
  `local_projection_io_error = 13` instead of the retryable ierr=3.
- Emit one parseable detail line per `reduce_hphi` recovery: propagator tag,
  `count_solv`, halved `mult_solv`, resulting phi point count.

All new stdout lines were checked against the campaign failure regex
(`run_reconstruction.py:17-21`) and the `Error in ripple_solver:` recovery
marker (line 22); none match, so failure detection and recovery counting are
unchanged.

## Preserved invariants

- With `NEO2_STAGE2_ATTRIBUTION_FILE` and `NEO2_PEAK_RSS_FILE` unset, solver
  behavior and outputs are bit-identical; runs without recoveries print no new
  lines.
- The join-end trace header and existing record kinds are unchanged; the new
  kinds are additive.
- CGS units, Fourier convention, discretizations, tolerances, ABI, and the
  serialized propagator layout are unchanged.

## Driver-side companion (flux_pumping repo)

- `run_reconstruction.py` `JOIN_END_TRACE_KINDS` must admit the four new kinds
  before `FLUX_PUMPING_JOIN_END_TRACE=1` runs against this head.
- The attribution campaign builds the footprint file from the stage-0
  `source_correction_p/m` rows for the closure-pair tags and sets
  `NEO2_STAGE2_ATTRIBUTION_FILE` for stage 2 only.

## Verification

Touched tests on this head:

```text
join_failure_diagnostic_test               PASS       0.06s
stage2_attribution_test                    PASS       0.06s
peak_rss_test                              PASS       0.11s
lorentz_projection_diagnostic_test         PASS       0.06s
propagator_io_test                         PASS       0.06s
crossing_geometry_padding_test             PASS       0.06s
qflux_interface_diagnostic_test            PASS       0.06s
```

Full suite comparison against the base head 6aec6fc (same worktree, same
toolchain): base `fo test --all` gives 28 passed / 16 failed, this head gives
30 passed / the identical 16 failures (pre-existing environmental
collision-operator and multispecies test failures, untouched by this PR; the
+2 are the new behavioral tests).

```text
$ fo
Build: OK
Tests: OK
Lint: OK
All stages passed (16.3s)
```

This PR is stacked on #165 (base `fix/flux-pumping/preserve-crossing-geometry`,
now including the crossing-scan bounds fix) and should remain unmerged until
owner review of the stack.

### DIFF
--- a/NEO-2-QL/CMakeSources.in
+++ b/NEO-2-QL/CMakeSources.in
@@ -12,6 +12,7 @@ set(NEO2_QL_SRC_FILES
   ripple_solver_ArnoldiOrder2_test.f90
   join_diagnostics_mod.f90
   lorentz_projection_diagnostics_mod.f90
+  stage2_attribution_mod.f90
   join_ripples_int.f90
   join_ripples.f90
   join_ends.f90
--- a/NEO-2-QL/CMakeSources.in
+++ b/NEO-2-QL/CMakeSources.in
@@ -13,6 +13,7 @@ set(NEO2_QL_SRC_FILES
   join_diagnostics_mod.f90
   lorentz_projection_diagnostics_mod.f90
   stage2_attribution_mod.f90
+  peak_rss_mod.f90
   join_ripples_int.f90
   join_ripples.f90
   join_ends.f90
--- a/NEO-2-QL/join_diagnostics_mod.f90
+++ b/NEO-2-QL/join_diagnostics_mod.f90
@@ -128,14 +128,24 @@ contains
         join_end_trace_enabled = allocated(join_end_output_filename)
     end function join_end_trace_enabled
 
+    !> Append one periodic-join compatibility record to the join-end trace.
+    !>
+    !> delta_eta_l/delta_eta_r are the per-band pitch measures of the joined
+    !> propagator boundaries used by the Lorentz solvability correction in
+    !> join_ends.f90; they are empty when that correction is inactive.  The
+    !> emitted source_correction_m/source_correction_p rows are the per-band,
+    !> per-force products delta_eta_l*facnorm and delta_eta_r*facnorm that
+    !> join_ends subtracts from source_m and source_p, so the periodic-closure
+    !> projection footprint is reconstructible band by band.
     subroutine record_join_end_compatibility(source_factor, source_before, &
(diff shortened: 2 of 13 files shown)


krystophny:

PR #168, opened 2026-07-18
Trace the flux-side closure correction and gate the Lorentz correction

krystophny:
Stacked on #167 (do not merge before it; owner review).

The periodic-join Lorentz solvability correction acts through two channels: the per-band source-row subtractions (traced since #167) and a per-force scalar subtracted uniformly from both flux extraction rows (untraced until now).

**What this adds**

1. **Flux-channel trace rows** in the join-end trace: `flux_factor`, `flux_before`, `flux_after`, `flux_scale` per force, plus per-band `flux_correction_m`/`flux_correction_p` rows in the same `tag,side,force,band` format as the source side. Bookkeeping note now recorded in the module doc: `join_ripples` composes flux rows only into qflux and other flux rows, so this channel can shift the extracted transport coefficients but structurally cannot affect the pointwise reconstruction (relevant to the converged periodic-interface jump measured on the pinned #167 campaigns: the flux channel is excluded a priori for the jump, and the trace makes its gamma-level effect auditable).

2. **`NEO2_DISABLE_JOIN_LORENTZ_CORRECTION=1`** as a controlled A/B switch: the periodic solve keeps the uncorrected sources and extraction rows while the trace still records the would-be correction factors; a loud audit line marks any disabled run (runners must require it, mirroring the stage-2 attribution marker). This is the decisive experiment for the interface jump after the chain-end source-row channel was excluded by the corrected attribution pass B (flux_pumping run record 19526225): if the jump vanishes with the correction disabled, it is the correction's distributed footprint on the periodic solution and not attributable to any single-propagator edit.

**Tests**: `test_join_failure_diagnostic` extended for the new recorder signature, the flux rows (uniform per-force scalar), a non-finite flux-scale rejection, and the default-off state of the switch; passes. Full local ctest: 30/30 environment-independent tests pass; the 16 `external-data regression` tests fail for a missing `NEO2_TEST_PATH` on this machine (pre-existing, unrelated).

No behavior change without the env switch: the correction arithmetic is untouched and the new trace rows appear only when `NEO2_JOIN_END_TRACE_FILE` is set.

### DIFF
--- a/NEO-2-QL/join_diagnostics_mod.f90
+++ b/NEO-2-QL/join_diagnostics_mod.f90
@@ -11,11 +11,41 @@ module join_diagnostics_mod
     integer, save :: join_end_sequence = 0
     logical, save :: normalization_output_initialized = .false.
     logical, save :: join_end_output_initialized = .false.
+    logical, save :: correction_switch_initialized = .false.
+    logical, save :: correction_switch_disabled = .false.
 
-    public :: join_end_trace_enabled, record_join_end_compatibility
+    public :: join_end_trace_enabled, join_lorentz_correction_disabled
+    public :: record_join_end_compatibility
     public :: report_join_failure, validate_join_normalization
 
 contains
+
+    !> True when NEO2_DISABLE_JOIN_LORENTZ_CORRECTION=1 requests the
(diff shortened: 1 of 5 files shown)


krystophny:

